PR #3, opened 2020-11-21
Fix ids

edalzell:
If you use the `id` column, there was an error when saving because the `generateId` call returned null and you can't pass null into a non-nullable column. You need to not send the `id` column value for `EntryModel` but include for `UuidEntryModel`.

Added `creating` hook so it's added every time a model is created, only on the UuidEntryModel.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         }
     },
     "require": {
-        "statamic/cms": "^3.0"
+        "statamic/cms": "dev-entries-with-int-id"
     },
     "require-dev": {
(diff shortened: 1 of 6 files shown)


edalzell:
@jasonvarga sorry to include the blueprint stuff in there, but I need to have all of this in one PR so it still works on our side (until you release, of course).

edalzell:
also, the composer require is mucked up cuz I need a specific (branched) Statamic. Will try to clean this up, somehow.

edalzell:
Splitting these up into separate PRs.

PR #5, opened 2020-11-23
Fix id

edalzell:
If you have `id` as an integer, you can't send null to it, as it throws an error.

To solve, we don't set the `id` in the creation, so that the DB can generate the correct one, when `EntryModel` is used.

When `UuidEntryModel` is used, we listen for the `creating` event and set the `id` to a generated UUID.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -26,7 +26,6 @@ class Entry extends FileEntry
         $class = app('statamic.eloquent.entries.model');
 
         return $class::findOrNew($this->id())->fill([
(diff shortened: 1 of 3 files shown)


edalzell:
@jasonvarga any chance we can get this merged in so I can stop using my branch?

jasonvarga:
I can't reproduce any error to begin with. Entries seem to be created without an issue.

Can you provide any other details? What DB are you using? MySQL? Postgres?

edalzell:
Postgres

jasonvarga:
Looks Postgres specific. Nice. Thanks.

PR #6, opened 2020-11-23
Add blueprint

edalzell:
The blueprint was not getting stored and it's needed so that you can set/change which blueprint is used on a particular entry.

Added to both migrations.


imbue:
Hey!

Would love to have this PR merged, is there anything I can do to help in the process?

PR #10, opened 2021-01-04
Support sub-queries

edalzell:
When you use sub-queries, `column` is a closure, and we want to pass it directly back.

### DIFF
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -29,7 +29,7 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
             $column = 'origin_id';
         }
 
-        if (! in_array($column, $this->columns)) {
+        if (! in_array($column, $this->columns) && is_string($column)) {
             $column = 'data->'.$column;
         }
 


jasonvarga:
Could you provide an example of what you're trying to achieve with this PR?

edalzell:
Sure, in a scope:

```
$query->where(function($subQuery) use($city) {
    ...do stuff in here w/ $subquery
});
```

Is that right @4dmind (Karol)?

jasonvarga:
Sorry, I meant the specific usage. But that's okay, not really necessary.

This will need to be an Eloquent-driver-specific feature since we can't support it on the file side any time soon.
So as long as you're using it in app code, and not an addon, you're fine.

edalzell:
> This will need to be an Eloquent-driver-specific feature since we can't support it on the file side any time soon.

Yes, of course, that's why this is in this repo?

jasonvarga:
Right but lets say an addon writes something like:

```
Entry::query()->where(function ($q) {
  $q->where(...);
});
```

The addon doesn't care if the user is using files or eloquent. But this code would only work for eloquent.

edalzell:
> The addon doesn't care if the user is using files or eloquent. But this code would only work for eloquent.

If there's a better way to handle this, I'm open to that. The better we can "merge" Statamic and Laravel the better.

jasonvarga:
The way is what you've done. Just saying it would have to be an Eloquent only feature to start with. I'm ok with that.

Maybe we can throw an exception if you pass a closure to the Stache builder saying 'not supported yet'.

edalzell:
> The way is what you've done. Just saying it would have to be an Eloquent only feature to start with. I'm ok with that. There's plenty of features in Laravel that are specific to drivers. e.g. [The "file" cache driver doesn't support tags.](https://laravel.com/docs/8.x/cache#cache-tags)
> 
> Maybe we can throw an exception if you pass a closure to the Stache builder saying 'not supported yet'.

That's a great idea. You want me to add it?

ryanmitchell:
I think this should be supported now that core query builders do it?

edalzell:
> I think this should be supported now that core query builders do it?

Maybe? I'll check when I get back to the project this is used on.

jasonvarga:
Closing in favor of #42

PR #20, opened 2021-05-28
Support all models

okaufmann:
For large applications where we have a lot of data we like to save it to the database as the repo will become to large otherwise (Several Gigabytes).

This Package did help a lot by providing a way to save the entities to the Database. Although, many inconsistencies arise when Taxonomies, Globals or Navigation links to entities IDs. The main problem is, that e.g. after changing the order of a page Collection, the tree is enriched with DB IDs. When pushing this code to a staging environment, other IDs exists and many pages are broken when Entries were not found.

For this reason, we extended the eloquent-driver to add support for all types of data.

If you have any feedback and like to have this PR in this package I will clean it and test it even more.

## TODO:
- [x] taxonomies
- [x] navigations
- [x] globals 
- [x] collections
- [ ] form submissions
- [ ] test multisite 
- [ ] cleanup
- [x] ensure new slug mechanism is working (https://github.com/statamic/cms/issues/714#issuecomment-856150217)

### DIFF
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Eloquent\Taxonomies;
 
+use Statamic\Facades\Site;
 use Statamic\Query\EloquentQueryBuilder;
 use Statamic\Taxonomies\TermCollection;
 
@@ -15,8 +16,13 @@ class TermQueryBuilder extends EloquentQueryBuilder
 
     protected function transform($items, $columns = [])
     {
-        return TermCollection::make($items)->map(function ($model) {
-            return Term::fromModel($model)->in($this->site);
+        $site = $this->site;
+        if(!$site) {
+            $site = Site::default()->handle();
+        }
+
+        return TermCollection::make($items)->map(function ($model) use($site) {
+            return Term::fromModel($model)->in($site);
         });
     }
 
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -7,6 +7,8 @@ use Statamic\Taxonomies\TermCollection;
 
 class TermQueryBuilder extends EloquentQueryBuilder
 {
+    protected $site = null;
+
     protected $columns = [
         'id', 'site', 'slug', 'uri', 'taxonomy', 'created_at', 'updated_at',
     ];
@@ -14,7 +16,7 @@ class TermQueryBuilder extends EloquentQueryBuilder
     protected function transform($items, $columns = [])
     {
         return TermCollection::make($items)->map(function ($model) {
-            return Term::fromModel($model);
+            return Term::fromModel($model)->in($this->site);
         });
     }
 
@@ -25,5 +27,16 @@ class TermQueryBuilder extends EloquentQueryBuilder
         }
 
         return $column;
+    }
+
+    public function where($column, $operator = null, $value = null)
+    {
+        if ($column === 'site') {
+            $this->site = $operator;
+
+            return $this;
+        }
+
+        return parent::where($column, $operator, $value);
     }
 }
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -1,0 +1,37 @@
+<?php
+
+
+namespace Statamic\Eloquent\Taxonomies;
+
+use Statamic\Eloquent\Entries\Entry;
+use Statamic\Entries\EntryCollection;
+use Statamic\Query\EloquentQueryBuilder;
+use Statamic\Taxonomies\TermCollection;
+
+
+class TermQueryBuilder extends EloquentQueryBuilder
+{
+    protected $columns = [
+        'id', 'site', 'slug', 'uri', 'taxonomy', 'created_at', 'updated_at',
+    ];
+
+    protected function transform($items, $columns = [])
+    {
+        return TermCollection::make($items)->map(function ($model) {
+            return Term::fromModel($model);
+        });
+    }
+
+    protected function column($column)
+    {
+        if ($column == 'origin') {
+            $column = 'origin_id';
+        }
+
+        if (! in_array($column, $this->columns)) {
+            $column = 'data->'.$column;
+        }
+
+        return $column;
+    }
+}
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -1,13 +1,9 @@
 <?php
 
-
 namespace Statamic\Eloquent\Taxonomies;
 
-use Statamic\Eloquent\Entries\Entry;
-use Statamic\Entries\EntryCollection;
 use Statamic\Query\EloquentQueryBuilder;
 use Statamic\Taxonomies\TermCollection;
-
 
 class TermQueryBuilder extends EloquentQueryBuilder
 {
@@ -24,10 +20,6 @@ class TermQueryBuilder extends EloquentQueryBuilder
 
     protected function column($column)
     {
-        if ($column == 'origin') {
-            $column = 'origin_id';
-        }
-
         if (! in_array($column, $this->columns)) {
             $column = 'data->'.$column;
         }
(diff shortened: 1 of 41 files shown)


jasonvarga:
Awesome.

jasonvarga:
I've marked this PR as a draft for you. Mark it as ready to review when you're ready. Thanks!

mbryne:
Looking forward to this being merged, cheers

okaufmann:
I'll work it out and make it ready as soon as possible, but I can't find any time atm though 😢

mbryne:
@okaufmann absolutely no worries mate, I completely understand. 


As I touched on in https://github.com/statamic/ideas/issues/312#issuecomment-849914523 we haven’t been able to deploy our Statamic installs to Vapor in over 6 months.

@jasonvarga is there anything you can do to get this over the line or do we need to wait for @okaufmann specifically?

We appreciate all the effort you have put in but just need a clear first-party solution here to Statamic on Vapor or at least a clear no.

mbryne:
@okaufmann I'm just writing to Statamic support about the issue, 

Are you open to them conducting their own review and integrating the changes from this pull request? Do you specifically want to go through that review process?

From our perspective, our purchasing decision was based on the promise of a fully functioning Eloquent driver so I'm asking them to integrate it internally but wanted to check in with you first.

okaufmann:
Hi @mbryne 

I started working on the eloquent-driver to be able to use Statamic for large amounts of data. I do not care how Statamic supports this. Either I finish this PR or it is supported out of the box 🙂

I'd love to hear that Statamic supports this natively, as the Magicians knows it even better and can implement it as performant and optimized as possible. But I also understand if the DB support is not wanted in the Core.

@jasonvarga What do you think? Is native DB support on the Roadmap for v3.2 or v4?

jasonvarga:
It will never be in the core, but since this is our addon it's as close to "native" as it's going to get.

From what I understood, you were happy to continue working on this PR. Once it was ready to be reviewed, we would review it, merge, and then the maintenance burden becomes ours.

mbryne:
@okaufmann no worries on that.

After speaking with Statamic support I offered to pick up where you left off with the review process if you didn't have time to do it. 

We are in a similar situation and migrating a lot of data that is not feasible to store in flat files, also the whole vapor thing.

It looks like you or (@benaja) are still in the process of making commits so I don't want to interfere with the great work you are both doing but I'd be happy to take the review process off your hands if that helps when you are done.

My understanding also is that this will always be an add-on and won't make it into core but I feel like its a very useful feature and keen to help where I can.

Daynnnnn:
Hey, I was planning on PR'ing into this repo, but by the time I'd got it to a decent point this PR had been already been opened, bad timing I guess 😅 I've also done an eloquent driver, and it's most of the way to finished. Mine admittedly doesn't seem as clean as this one, and I've done some stuff differently, but if there's anything you want to rip out of mine to get this over the line, or there's anything I could help with, let me know. I'd love to have a 1st party solution to file-less statamic.

https://github.com/Daynnnnn/statamic-database

MerlinBrasil:
Any idea when this will be completed and ready for testing?

This is on my critical path and would be happy to beta test if needed.

Without this, we'd have to drop Statamic and kinda start over!

Profound thanks for your good work thus far.

Best regards,
Merlin

src/Taxonomies/TermQueryBuilder.php
```diff
@@ -15,8 +16,13 @@ class TermQueryBuilder extends EloquentQueryBuilder
 
     protected function transform($items, $columns = [])
     {
-        return TermCollection::make($items)->map(function ($model) {
-            return Term::fromModel($model)->in($this->site);
+        $site = $this->site;
+        if(!$site) {
+            $site = Site::default()->handle();
+        }
+
```
edalzell, line 23: would this be nicer as

```php
$site = $this->site ?? Site::default()->handle();
```

rickmills:
I've been testing this out on a project and _may_ have found a bug, but unsure at this stage if it's specific to the eloquent driver. It could do with another test.

Using a tree structure collection, I have the following entries:

- Parent 1 (Blueprint 1)
    - Child 1 (Blueprint 2)
    - Child 2 (Blueprint 2)
- Parent 2 (Blueprint 1)
    - Child 3 (Blueprint 2)

These save as expected. However if you move, for example `Child 2` into `Parent 2` thats where problems start. The `trees` table correctly updates and shows the new tree structure in the control panel, and the URL's update.

However on the actual `Child 2` entries db record, the `parent` value inside the `data` column json array remains being set to Parent 1. The consequence of this is that any in-page references are then broken. So if for example you want to list all your Child records inside the Parent template, Parent 1 as a page would show Child 2 even though it was no longer part of that group.

I thought the issue was that `CollectionTreeSaved` event was being ignored so wasn't triggering an update of the parent on the entry, however if you go into the entry itself in the admin area and update the parent there it's also not updating the entry record, so perhaps something deeper going on.

crackAT:
Hei all, 

can i already use this draft for my production site? Because we are waiting for taxonomy database support and i think this is already done in the draft? 

Thanks for the great work btw.

crackAT:
Hi all, 

i tried to use this great feature and got a bug on composer update. Maybe this is a known issue: 

```
 Symfony\Component\ErrorHandler\Error\FatalError 

  Declaration of Statamic\Eloquent\Taxonomies\TermRepository::find($id): ?Statamic\Taxonomies\LocalizedTerm must be compatible with Statamic\Stache\Repositories\TermRepository::find($id): ?Statamic\Contracts\Taxonomies\Term

  at vendor/statamic/eloquent-driver/src/Taxonomies/TermRepository.php:50
     46▕         return [
     47▕             TermContract::class => Term::class,
     48▕         ];
     49▕     }
  ➜  50▕ }
     51▕ 
```

Installed it on a fresh knew statamic installation.

Statamic 3.2.29 Pro
Laravel 8.78.0
PHP 7.3.29-to-be-removed-in-future-macOS

rickmills:
> Hi all,
> 
> i tried to use this great feature and got a bug on composer update. Maybe this is a known issue:
>
> ...
> 
> Statamic 3.2.29 Pro
> Laravel 8.78.0
> PHP 7.3.29-to-be-removed-in-future-macOS

Yes it looks like an update to the Stache TermRepository file has caused a couple of bits to stop working on the latest version. 

That aside you'll also need to be on at lease PHP 8.0 I believe, as theres some 8.0+ code in this branch.

crackAT:
Hi, thanks @rickmills for the php8 hint, now its working. 

Everything is working very well and the cms is much quicker while using a lot of taxonomies. 
one error occured when entering a collection and the blueprint is set to hidden. Maybe a styled error message would be better than a laravel error :) 

One feature request from my side: Storing the image datas in the database would also be great. Datas like, size, width, height, filename, uri and others.

crackAT:
After using the eloquent driver for Taxonomy's i have some, maybe relevant notes: 
- TaxonomyTerm fields are using the slug and not the Tax-title. 
- TaxonomyTerm slugs should be unique. Atm the are unique per taxonomy but in the api response i'm only receiving the term slug and with multiple equal slugs it's a little bit difficult do find the right slug with the right taxonomy type.

ryanmitchell:
I hope you don't mind me chipping in here, but I have taken this work and extended it to apply to forms, form submissions and asset containers which you can see here: https://github.com/ryanmitchell/eloquent-driver/tree/fixes-to-all-models-pr

I can't push changes to this PR, otherwise I would add them here.

Beyond the additions I have refactored the code considerably:

* Config file is grouped and rationalised
* Used contracts where available
* Ensured we are always using the bound models
* Allowed for table names to be prefixed via a config setting
* Restructured the collections table to avoid having to keep adding new columns
* Support the collection new title format option in Statamic 3.2
* Re-instated migrations (and updated the docs accordingly)

From what I can see this is everything now working from the database with the exception of blueprints, which I'm happy to work on if its felt to be useful.

ryanmitchell:
If anyone knows how to get tests working for this PR, and is willing to lend a hand writing them that would be super helpful.

ryanmitchell:
@okaufmann i raised a PR against your fork - can you check it and merge it in so we can move this PR to the next stage?

jasonvarga:
Thank you for this wonderful contribution. Since Ryan has extended on it over at #42, we'll continue over there.

PR #40, opened 2022-01-23
Bug fix: Make sure we select all columns

ryanmitchell:
With the recent changes on core (https://github.com/statamic/cms/pull/5068) this driver no longer works with collection tags (as reported here https://github.com/statamic/cms/issues/5108). The quick fix is to use a eloquent-driver specific select() that ensures we still get all columns.

It would be nice to make the select() actually work so only the appropriate keys from the data column are returned, but for now this feels better than it not working.


devingray (approved):
Clean

jasonvarga:
Thanks, but I'll fix it on the Statamic side. Hang tight, sorry about that!

PR #41, opened 2022-02-17
Support all models

juliomenendez:
Takes the work on PR #20 by @okaufmann and adds the rest of the models to make Statamic fully independent from local storage.


ryanmitchell:
... i think we have conflicting PRs :) 

I took a lot of what you did and reconfigured it to make it extendable and use contracts etc... I was waiting on @okaufmann agreeing to merge it to his to make it easier. I've also added other models and written console imports.

Can we close this and join forces to give it more change of getting merged?

ryanmitchell:
Here's my branch by the way - https://github.com/ryanmitchell/eloquent-driver/tree/fixes-to-all-models-pr

I only noticed yours last night after I had implemented the models and restructured it all (!), I wish I had seen it at the start as it would have made it a much quicker job. I ended up taking your blueprints and field sets and merging them in too.

Theres quite a difference in code and fixes that I've put in place so please check it out.

juliomenendez:
@ryanmitchell sorry for the delay replying, I've been crazy busy this last week. I'll review your changes to see if we can use your implementation but one thing we do need on our implementation is the full user roles and groups in the database and for the time being, it doesn't look like it's in the works by Statamic.

ryanmitchell:
No problem. If Jason wants it in this driver then I’ll bring it over from your work - but it makes more sense to me to be part of the core eloquent user support.

jasonvarga:
Thank you for the great contribution. Since Ryan's extended it over on #42, we'll continue there.

PR #48, opened 2022-05-30
Add a default `lastModified`

simonhamp:
Useful when you're duplicating an entry and the new entry has a new model and therefore doesn't have a `lastModified` value


simonhamp:
Tests fix is in https://github.com/statamic/eloquent-driver/pull/49

FrittenKeeZ:
Pretty sure this PR is obsolete now.

jasonvarga:
Yup.

PR #42, opened 2022-02-17
Support all models

ryanmitchell:
Sorry for the duplication of PRs here but I have reached out to the authors of both the previous PRs to try and combine efforts but have had no response from either.

As I indicated [here](https://github.com/statamic/eloquent-driver/pull/20#issuecomment-1040171068) I have taken the existing PR (#20) and added the rest of the models. I subsequently discovered the work that has since been opened in #41 and have added the additional blueprints and field sets work from it.

~~I have intentionally not included user groups and role as IMO they should be handled by Statamic core as part of the eloquent users setup.~~ **Update:** I've opened a PR in statamic/cms for this: https://github.com/statamic/cms/pull/5686

So to confirm this PR includes:

- [x] taxonomies
- [x] navigations
- [x] globals
- [x] collections
- [x] form submissions
- [x] asset containers
- [x] assets meta
- [x] blueprints
- [x] field sets 
- [x] revisions 

I have also refactored the code allow for devs to opt in via config to each 'model', I have ensured it uses the bound models, and I have written console import tasks to allow migration from file based data to eloquent.

Finally I have applied fixes from any open PRs and ensured they are included (with the exception of Jason's).

@jasonvarga I'd appreciate it if you could review this and let me know what you think.

PR #52, opened 2022-08-05
Automatically run php cs fixer

okaufmann:
Hi

This PR simply adds a Github Action which will run php-cs-fixer on push.


FrittenKeeZ:
When #42 gets merged it would make sense to switch to Laravel Pint
https://laravelmagazine.com/blog/run-laravel-pint-as-part-of-your-ci-pipeline-with-github-actions

jasonvarga:
I'd rather use StyleCI like we do on statamic/cms anyway. Thank you though.

okaufmann:
Nice, the main thing is that the code is formatted now. Thank you.

PR #53, opened 2022-09-02
force the column returns with table name as alias

maxwkf:
it helps to resolve the problem of same column name overrided by a joint table

### DIFF
--- a/tests/Data/Entries/EntryQueryBuilderTest.php
+++ b/tests/Data/Entries/EntryQueryBuilderTest.php
@@ -690,7 +690,7 @@ class EntryQueryBuilderTest extends TestCase
     }
 
     /** @test */
-    public function entries_retrieval_with_join_table()
+    public function entries_can_be_retrieved_and_are_not_null_when_using_join()
     {
         Collection::make('posts')->save();
         EntryFactory::id('1')->slug('post-1')->collection('posts')->data(['title' => 'Post 1', 'author' => 'John Doe', 'location' => 4])->create();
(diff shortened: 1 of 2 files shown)


tests/Data/Entries/EntryQueryBuilderTest.php
```diff
@@ -688,4 +688,42 @@ public function entries_are_found_using_offset()
         $this->assertCount(2, $entries);
         $this->assertEquals(['Post 2', 'Post 3'], $entries->map->title->all());
     }
+    
+    /** @test */
+    public function entries_retrieval_with_join_table()
```
ryanmitchell, line 693: maybe rename it to something more descriptive like `entries_can_be_retrieved_and_are_not_null_when_using_join`

jasonvarga:
Sorry for the delay on this.

I'm more than happy to have this addressed, but joins are really only working by fluke because we proxy methods onto the eloquent builder. Joins aren't used anywhere else in Statamic yet.

Could you please provide a real use case for this, or more clearly explain the issue?

Also, please fix the merge conflicts.

Thanks!

maxwkf:
> Sorry for the delay on this.
> 
> I'm more than happy to have this addressed, but joins are really only working by fluke because we proxy methods onto the eloquent builder. Joins aren't used anywhere else in Statamic yet.
> 
> Could you please provide a real use case for this, or more clearly explain the issue?
> 
> Also, please fix the merge conflicts.
> 
> Thanks!

Sorry for the late reply.

I think the conversation started with Ryan from here
https://discord.com/channels/489818810157891584/489819906540568593/1015191243183112212.

This conversation goes to a thread at last (https://discord.com/channels/489818810157891584/1015194102515638282).

As I need to create my own tag for searching entries which related to other collection entries, I need to create my custom search query.

PR #54, opened 2022-09-02
make command register easier with ClassMapGenerator

enespolat24:
Hi , i refactored command discovery in ServiceProvider. now we don't need to add them manually


jasonvarga:
Thank you, that's pretty cool but I don't see the benefit of the extra overhead running the classmap generator on every request, where an array is simple enough.

PR #56, opened 2022-09-05
Fix table name usage in clauses

### LINKED ISSUE #55: Where on join query with alias not work
# Where on join query with alias not work
When a where clause apply on join query with alias, it cannot parse the correct where.

Let us consider the following query.  

__PHP Query__
```PHP
$query = Entry::query()
  // join self as e
  ->join('entries as e',fn($join) => $join
  ->whereColumn('e.id', 'entries.id')
  ->where('e.collection', 'properties')
  ->where('e.published', 1)
);
// join locations
$query->leftJoin('entries as locations', function($join) {
    $join
        ->where('locations.collection', 'locations')
        ->on('locations.id', 'e.data->location')
        ;
});
// where on joint clause
$query
  // this is expected to search on the JSON data field for property_type
  ->where('e.data->property_type', 'lodge')
  // this is expected to search on the locations slug for value 'shaldon'
  ->where('locations.slug', 'shaldon')
;
```

__SQL Generated__

The where property_type putting the e.data in the json key field but not the data field
```SQL
select * from `entries`
  inner join `entries` as `e`
    on `e`.`id` = `entries`.`id` and `e`.`collection` = 'properties' and `e`.`published` = '1'
  left join `entries` as `locations`
    on `locations`.`collection` = 'locations' and `locations`.`id` = json_unquote(json_extract(`e`.`data`, '$."location"'))
  where
    json_unquote(json_extract(`data`, '$."e.data"."property_type"')) = 'lodge' // Wrong SQL
    and json_unquote(json_extract(`data`, '$."locations.slug"')) = 'shaldon' // Wrong SQL
```
__SQL Expected__
```SQL
select * from `entries`
  inner join `entries` as `e`
    on `e`.`id` = `entries`.`id` and `e`.`collection` = 'properties' and `e`.`published` = '1'
  left join `entries` as `locations`
    on `locations`.`collection` = 'locations' and `locations`.`id` = json_unquote(json_extract(`e`.`data`, '$."location"')) 
  where
    json_unquote(json_extract(`e`.`data`, '$."property_type"')) = 'lodge' // SQL expected
    and `locations`.`slug` = 'shaldon' // SQL expected
```
## Cause
The problem caused by the column check in `src/Entries/EntryQueryBuilder.php`.

```PHP
protected function column($column)
{
    if ($column == 'origin') {
        $column = 'origin_id';
    }

    if (! in_array($column, self::COLUMNS)) {
        $column = 'data->'.$column;
    } else {
        $column = parent::column($column);
    }

    return $column;
}
```

maxwkf:
Fixes #55

### DIFF
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -18,8 +18,6 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         'date', 'collection', 'created_at', 'updated_at',
     ];
 
-    protected $isApplyColumnCheck = true;
-
     protected function transform($items, $columns = [])
     {
         $items = EntryCollection::make($items)->map(function ($model) {
@@ -34,20 +32,21 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         if (! is_string($column)) {
             return $column;
         }
+        
+        $table = Str::contains($column, '.') ? Str::before($column, '.') : '';
+        $column = Str::after($column, '.');
 
         if ($column == 'origin') {
             $column = 'origin_id';
         }
 
-        if ($this->isApplyColumnCheck() && ! in_array($column, self::COLUMNS)) {
-            if (! Str::startsWith($column, 'data->')) {
+        if (! in_array($column, self::COLUMNS)) {
+            if (! Str::contains($column, 'data->')) {
                 $column = 'data->'.$column;
             }
-        } else {
-            $column = parent::column($column);
         }
 
-        return $column;
+        return ($table ? $table.'.' : '').$column;
     }
 
     public function find($id, $columns = ['*'])
@@ -79,17 +78,5 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         $this->addTaxonomyWheres();
 
         return parent::count();
-    }
-
-    public function setApplyColumnCheck($value = true)
-    {
-        $this->isApplyColumnCheck = $value;
-
-        return $this;
-    }
-
-    public function isApplyColumnCheck($value = null)
-    {
-        return $this->isApplyColumnCheck;
     }
 }
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -18,6 +18,8 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         'date', 'collection', 'created_at', 'updated_at',
     ];
 
+    protected $isApplyColumnCheck = true;
+
     protected function transform($items, $columns = [])
     {
         $items = EntryCollection::make($items)->map(function ($model) {
@@ -37,10 +39,12 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
             $column = 'origin_id';
         }
 
-        if (! in_array($column, self::COLUMNS)) {
+        if ($this->isApplyColumnCheck() && ! in_array($column, self::COLUMNS)) {
             if (! Str::startsWith($column, 'data->')) {
                 $column = 'data->'.$column;
             }
+        } else {
+            $column = parent::column($column);
         }
 
         return $column;
@@ -75,5 +79,14 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         $this->addTaxonomyWheres();
 
         return parent::count();
+    }
+
+    public function setApplyColumnCheck($value = true) {
+        $this->isApplyColumnCheck = $value;
+        return $this;
+    }
+
+    public function isApplyColumnCheck($value = null) {
+        return $this->isApplyColumnCheck;
     }
 }
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -41,7 +41,7 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         }
 
         if (! in_array($column, self::COLUMNS)) {
-            if (! Str::contains($column, 'data->')) {
+            if (! Str::startsWith($column, 'data->')) {
                 $column = 'data->'.$column;
             }
         }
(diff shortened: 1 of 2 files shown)


maxwkf:
By using the function setApplyColumnCheck, it then bypass the checking on column and apply original where clause.
```php
$query = Entry::query()
        ->setApplyColumnCheck(false)
```

ryanmitchell:
Would a better fix not be to change the `!Str::startsWith('data->'` to be `Str::contains('data->` ?

maxwkfcc:
> Str::startsWith('data->'

This line is not introduced by me.

maxwkfcc:
> contains

Do you mean like this?
```php
    protected function column($column)
    {
        if (! is_string($column)) {
            return $column;
        }

        if ($column == 'origin') {
            $column = 'origin_id';
        }

        if (in_array($column, self::COLUMNS)) {
            if (! Str::contains($column, 'data->')) {
                $column = 'data->'.$column;
            }
        }

        return $column;
    }

```

ryanmitchell:
Yep - basically... one small change (maybe a copy/paste issue on your side).


```
    protected function column($column)
    {
        if (! is_string($column)) {
            return $column;
        }

        if ($column == 'origin') {
            $column = 'origin_id';
        }

        if (! in_array($column, self::COLUMNS)) {
            if (! Str::contains($column, 'data->')) {
                $column = 'data->'.$column;
            }
        }

        return $column;
    }
```

does that change fix your issue?

maxwkfcc:
> startsWith

It looks you are right.  Let me check it up.  Thanks

maxwkf:
> Yep - basically... one small change (maybe a copy/paste issue on your side).
> 
> ```
>     protected function column($column)
>     {
>         if (! is_string($column)) {
>             return $column;
>         }
> 
>         if ($column == 'origin') {
>             $column = 'origin_id';
>         }
> 
>         if (! in_array($column, self::COLUMNS)) {
>             if (! Str::contains($column, 'data->')) {
>                 $column = 'data->'.$column;
>             }
>         }
> 
>         return $column;
>     }
> ```
> 
> does that change fix your issue?

This would generate the following SQL in which the location part is incorrect.
```SQL
select * from "entries" inner join "entries" as "e" on "e"."id" = "entries"."id" and "e"."collection" = ? left join "entries" as "locations" on "locations"."collection" = ? and "locations"."id" = json_extract("e"."data", '$."location"') where json_extract("e"."data", '$."title"') like ? and json_extract("data", '$."locations.slug"') = ?
```

ryanmitchell:
What about adding a check for the table alias... something like:

```
    protected function column($column)
    {
        if (! is_string($column)) {
            return $column;
        }
        
        $table = Str::contains($column, '.') ? Str::before($column, '.') : '';
        $column = Str::after($column, '.');

        if ($column == 'origin') {
            $column = 'origin_id';
        }

        if (! in_array($column, self::COLUMNS)) {
            if (! Str::contains($column, 'data->')) {
                $column = 'data->'.$column;
            }
        }

        return ($table ? $table.'.' : '').$column;
    }
```

maxwkf:
> What about adding a check for the table alias... something like:
> 
> ```
>     protected function column($column)
>     {
>         if (! is_string($column)) {
>             return $column;
>         }
>         
>         $table = Str::contains($column, '.') ? Str::before($column, '.') : '';
>         $column = Str::after($column, '.');
> 
>         if ($column == 'origin') {
>             $column = 'origin_id';
>         }
> 
>         if (! in_array($column, self::COLUMNS)) {
>             if (! Str::contains($column, 'data->')) {
>                 $column = 'data->'.$column;
>             }
>         }
> 
>         return ($table ? $table.'.' : '').$column;
>     }
> ```

Thanks a lot.  You logic is right.  Do you think this is simplier?
```php
    protected function column($column)
    {
        if (! is_string($column)) {
            return $column;
        }

        if ($column == 'origin') {
            $column = 'origin_id';
        }

        if (! in_array($column, self::COLUMNS)) {
            if (! Str::contains($column, 'data->')) {
                if (! Str::contains($column, '.') ) {
                    $column = 'data->'.$column;
                    
                }
            }
        }

        return $column;
    }
```

ryanmitchell:
Simpler yes but it wouldn’t work the same. Stick with what I did :)

maxwkf:
> Simpler yes but it wouldn’t work the same. Stick with what I did :)
> […](#)
> On 5 Sep 2022, at 14:49, maxwkf ***@***.***> wrote: ﻿ What about adding a check for the table alias... something like: protected function column($column) { if (! is_string($column)) { return $column; } $table = Str::contains($column, '.') ? Str::before($column, '.') : ''; $column = Str::after($column, '.'); if ($column == 'origin') { $column = 'origin_id'; } if (! in_array($column, self::COLUMNS)) { if (! Str::contains($column, 'data->')) { $column = 'data->'.$column; } } return ($table ? $table.'.' : '').$column; } Thanks a lot. You logic is right. Do you think this is simplier? protected function column($column) { if (! is_string($column)) { return $column; } if ($column == 'origin') { $column = 'origin_id'; } if (! in_array($column, self::COLUMNS)) { if (! Str::contains($column, 'data->')) { if (! Str::contains($column, '.') ) { $column = 'data->'.$column; } } } return $column; } — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you commented.

Hi Ryan,

Thanks for your help.  Updated according to our discussion.  Do I need to close the pull request or leave it as open?

ryanmitchell:
Leave it open :)

src/Entries/EntryQueryBuilder.php
```diff
@@ -33,17 +33,20 @@ protected function column($column)
...
+        $table = Str::contains($column, '.') ? Str::before($column, '.') : '';
+        $column = Str::after($column, '.');
+
         if ($column == 'origin') {
             $column = 'origin_id';
         }
 
         if (! in_array($column, self::COLUMNS)) {
-            if (! Str::startsWith($column, 'data->')) {
+            if (! Str::contains($column, 'data->')) {
```
jasonvarga, line 44: The test still passes if I revert this back to `startsWith`. Is it necessary to change it?

ryanmitchell: Yeah should be starts with - it’s a better approach

jasonvarga: I agree. It seems like maybe the `contains` check just stuck around accidentally while you figured out the solution to the issue?

ryanmitchell: Yeah - contains could potentially return false results too. Please change it :)

PR #58, opened 2022-09-05
Apply selectedQueryColumns on transform too

ryanmitchell:
Noticed this was missing while working on #57


jasonvarga:
This also fixes an issue where in the REST API if you use `?fields=foo,bar` it would still show all fields.
With this PR it only shows those two fields as expected.

PR #60, opened 2022-09-15
prevent crated_at being null while submitting a from submission

enespolat24:
Hi , i was tinkering with form submission then i noticed that created_at being null. Can you review my changes to see if it fixes the problem. thank you 

image

### DIFF
--- a/src/Forms/Submission.php
+++ b/src/Forms/Submission.php
@@ -28,7 +28,7 @@ class Submission extends FileEntry
 
         $model = $class::findOrNew($this->id());
 
-        return (!empty($model->id)) ? $model->fill([
+        return (! empty($model->id)) ? $model->fill([
             'data' => $this->data,
         ]) : $model->fill([
             'data' => $this->data,
--- a/src/Forms/Submission.php
+++ b/src/Forms/Submission.php
@@ -6,6 +6,7 @@ use Statamic\Eloquent\Forms\SubmissionModel as Model;
 use Statamic\Events\SubmissionDeleted;
 use Statamic\Events\SubmissionSaved;
 use Statamic\Forms\Submission as FileEntry;
+use Illuminate\Support\Carbon;
 
 class Submission extends FileEntry
 {
--- a/src/Forms/Submission.php
+++ b/src/Forms/Submission.php
@@ -2,6 +2,7 @@
 
 namespace Statamic\Eloquent\Forms;
 
+use Illuminate\Support\Carbon;
 use Statamic\Eloquent\Forms\SubmissionModel as Model;
 use Statamic\Events\SubmissionDeleted;
 use Statamic\Events\SubmissionSaved;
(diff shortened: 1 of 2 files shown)


src/Forms/Submission.php
```diff
@@ -52,7 +52,7 @@ public function date($date = null)
             $this->date = $date;
         }
 
-        return $this->date;
+        return $this->date = Carbon::now();
```
ryanmitchell, line 55: Would be better as `$this->date ?? ($this->date = Carbon::now())` so that it doesnt overwrite the existing value

enespolat24:
thank you for your help @ryanmitchell

enespolat24:
this pr should  fix #61

andrasvati:
Hi, beside the modifications you've made I'd suggest to make the firstOrNew attribute based on the id, not on form_id and timestamp, to prevent parallel submissions with the same timestamp to overwrite each other. 
`public function toModel()
    {
        $class = app('statamic.eloquent.forms.submission_model');
        $timestamp = (new $class)->fromDateTime($this->date());

        **$model = $class::findOrNew($this->id());
        return (!empty($model->id)) ? $model->fill([
            'data' => $this->data,
        ]) : $model->fill([
            'data' => $this->data,
            'form_id' => $this->form->model()->id,
            'created_at' => $timestamp,
        ]);**
    }`

okaufmann:
This breaks all made submissions when they are shown in statamic and is therefore a critical bug. 

@jasonvarga can you take a look and merge if possible?

enespolat24:
@okaufmann i made some changes. I don't fully understand the overwrite situation. Do you have a chance to send an example case or video?

Thank you for your help !

enespolat24:
i also need help about tests

ryanmitchell:
@enespolat24 I have the tests working... can you add me to your fork so I can push to it?

okaufmann:
@enespolat24 Sorry for the delay.

> I don't fully understand the overwrite situation. Do you have a chance to send an example case or video?
In the current state, the date field will not be saved. Therefore, a form can only have one submission (as the filter is form_id and created_at, where created_at is always null).

@ryanmitchell Is there anything I could help? I have time now

ryanmitchell:
@okaufmann i added a test here - seems to work ok for me: https://github.com/statamic/eloquent-driver/blob/cc3780c3a9654346206cb1de4ba96a9c181ca234/tests/Forms/FormSubmissionTest.php#L56

okaufmann:
@ryanmitchell I would also freeze the time, so it creates two submissions width the same timestamp. So we can be sure it does not overwrite any submission.

ryanmitchell:
@okaufmann that will fail due to the unique constraint in the database schema. Two submissions should never have the same microsecond time...

okaufmann:
@ryanmitchell You're right. Then the tests cover all cases 💪

enespolat24:
@jasonvarga can you check if this is mergeable. This fixes a serious problem we face

Issue #63, opened 2022-10-04
`fresh()` not returning new instance


helloiamlukas:
#### Description
If you are editing an entry, save it and try to get a new instance with the `fresh()` method you will get the outdated entry back.

#### Example

```php
$entry->set('title', 'New Title');
$entry->save();

dd(
    $entry->fresh()->title,
    \Statamic\Facades\Entry::query()->where('id', $entry->id)->first()->title,
);
```

This will output two different values.

helloiamlukas:
The reason for this problem is the blink cache in the `find()` method.

https://github.com/statamic/eloquent-driver/blob/2c4a775accb30803120b009209bc80ebcca23093/src/Entries/EntryRepository.php#L20-L25

ryanmitchell:
I've opened a PR here: https://github.com/statamic/eloquent-driver/pull/64

helloiamlukas:
That was fast! I just wanted to submit a PR too, but I guess you were faster :) Thanks @ryanmitchell 🚀

ryanmitchell:
@helloiamlukas I feel responsible :)

PR #64, opened 2022-10-04
Store entry to blink cache on save

ryanmitchell:
Fixes: #63

### DIFF
--- a/src/Entries/EntryRepository.php
+++ b/src/Entries/EntryRepository.php
@@ -55,6 +55,8 @@ class EntryRepository extends StacheRepository
         $model->save();
 
         $entry->model($model->fresh());
+
+        Blink::put("eloquent-entry-{$entry->id()}", $entry);
     }
 
     public function delete($entry)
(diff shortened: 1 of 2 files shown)


src/Entries/EntryRepository.php
```diff
@@ -55,6 +55,8 @@ public function save($entry)
         $model->save();
 
         $entry->model($model->fresh());
+
+        Blink::put("eloquent-entry-{$entry->id()}", $entry);
```
jasonvarga, line 59: Probably want to also do it for `"eloquent-entry-{$entry->uri()}"` as per https://github.com/statamic/eloquent-driver/blob/2c4a775accb30803120b009209bc80ebcca23093/src/Entries/EntryRepository.php#L38-L41

jasonvarga (changes requested):

src/Entries/EntryRepository.php
ryanmitchell: Yep - thats done

Issue #67, opened 2022-10-04
Postgres: Can't create entry


helloiamlukas:
#### Description
When using Postgres, the following error will be thrown when trying to create a new entry:
```
SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "id" violates not-null constraint
```

The driver tries to use `null` as value for the `id` column for new entries.
https://github.com/statamic/eloquent-driver/blob/2c4a775accb30803120b009209bc80ebcca23093/src/Entries/Entry.php#L44-L46

Can we just delete this line, so that the `id` will not be written? Or is there any case where we need the `id` to be there?

edalzell:
@ryanmitchell is the fix above the right one?

ryanmitchell:
Yeah I think its fine to drop if from the fill, its picked up by the findOrNew()

PR #68, opened 2022-10-04
Only fill entry model with id if it already has one, to prevent postgres error

helloiamlukas:
fixes https://github.com/statamic/eloquent-driver/issues/67

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": "^8.0",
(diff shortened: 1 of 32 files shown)


edalzell:
@helloiamlukas can you please resolve the merge issues?

jasonvarga:
FYI currently this reverts the fix from #19. If you were to import entries, they'd lose their existing IDs.

I'll update this PR appropriately.

PR #71, opened 2022-10-07
Delete old meta after move/rename

### LINKED ISSUE #70: Asset meta data will not be updated when file is moved
## Issue
Moving an asset into different folder results in a duplicate entry + missing the meta data.

CleanShot 2022-10-07 at 15 23 31@2x

ryanmitchell:
Closes #70 #80


leganz:
Will this also cover the issue for renaming assets?

src/Assets/Asset.php
```diff
@@ -89,18 +91,44 @@ public function writeMeta($meta)
...
+        $oldMetaPath = $this->metaPath();
+        $newPath = Str::removeLeft(Path::tidy($folder.'/'.$filename.'.'.pathinfo($oldPath, PATHINFO_EXTENSION)), '/');
+
+        $this->hydrate();
+        $this->disk()->rename($oldPath, $newPath);
+        $this->path($newPath);
+        $this->save();
+
+        if ($oldMetaPath != $this->metaPath()) {
+            if ($oldMetaPath = app('statamic.eloquent.assets.model')::whereHandle($this->container()->handle().'::'.$oldMeta)->first()) {
```
leganz, line 105: That will throw an exception "$oldMeta is not defined".

leganz:
As mentioned in #80 - the writeMeta data will be called 3 times. only at the first time it has the metadata

leganz:
@ryanmitchell  

If this line:

https://github.com/statamic/eloquent-driver/blob/de33af216ca369a3c4ccb07aa97fa176b7b97dc9/src/Assets/Asset.php#L63

is not set meta data won't be overwritten. But i think we could change the code to:

```php 
$handle = $this->container()->handle().'::'.$this->metaPath();
        if ($model = app('statamic.eloquent.assets.model')::where('handle', $handle)->first()) {
            $meta = $model->data;
        }
        else {
            $meta = $this->generateMeta();
        }   
```

ryanmitchell:
Hmm that feels like an unnecessary db query - keen to avoid that.
Does[ this change help](https://github.com/statamic/eloquent-driver/pull/71/commits/e7690c8f76cf8dc6843f8a43fb27322f3cf40714)?

leganz:
> Hmm that feels like an unnecessary db query - keen to avoid that. Does[ this change help](https://github.com/statamic/eloquent-driver/pull/71/commits/e7690c8f76cf8dc6843f8a43fb27322f3cf40714)?

I am afraid it doesn't solve the issue right now. :/

leganz:
@ryanmitchell  do we really need "$this->writeMeta($meta = $this->generateMeta());" in metaValue? I just removed it from the method (your latest version) and moving/renameing worked without losing meta data. eg:

```php 
private function metaValue($key)
    {
        $value = Arr::get($this->meta(), $key);

        if (! is_null($value)) {
            return $value;
        }

        Cache::forget($this->metaCacheKey());

        $meta = $this->generateMeta();

        //$this->writeMeta($meta = $this->generateMeta());

        return Arr::get($meta, $key);
    }
```

ryanmitchell:
Im not sure about that - its in the file based version so it must be there for a reason.
I cant recreate your issues so it would be helpful to see a sample repository so I can see what you are seeing.

leganz:
I am going to prepare one :)

ryanmitchell:
@leganz I've updated this and refactored a bit so it now works with the latest versions of Statamic and this driver. Do you want to review it before I merge it in?

PR #82, opened 2022-10-25
Use `jsonb` data type for PostgreSQL to increase perfomance

helloiamlukas:
To improve perfomance for Postgres databases, the `json` columns can use the `jsonb` data type.

This is will reduce responsetimes drastically when dealing with large datasets.

### DIFF
--- a/database/migrations/create_navigations_table.php.stub
+++ b/database/migrations/create_navigations_table.php.stub
@@ -11,7 +11,6 @@ return new class extends Migration {
             $table->id();
             $table->string('handle')->unique();
             $table->string('title');
-            $table->string('locale')->nullable()->index();
             $table->jsonb('settings')->nullable();
             $table->timestamps();
         });
--- a/database/migrations/create_navigations_table.php.stub
+++ b/database/migrations/create_navigations_table.php.stub
@@ -11,7 +11,12 @@ return new class extends Migration {
             $table->id();
             $table->string('handle')->unique();
             $table->string('title');
-            $table->json('settings')->nullable();
+            $table->string('locale')->nullable()->index();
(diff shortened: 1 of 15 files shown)


jasonvarga (changes requested):
I'm fairly sure you can just to `$table->jsonb('name')` without all those conditions.

On DBs that don't support jsonb, it'll fall back to json or text.

Postgres uses `jsonb` as expected:

https://github.com/laravel/framework/blob/ec3958de050d14f15832aa7604fb6834efa807bf/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php#L780-L783


MySQL goes to `json`:

https://github.com/laravel/framework/blob/ec3958de050d14f15832aa7604fb6834efa807bf/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php#L718-L721

SQLite goes to `text`:

https://github.com/laravel/framework/blob/ec3958de050d14f15832aa7604fb6834efa807bf/src/Illuminate/Database/Schema/Grammars/SQLiteGrammar.php#L624-L633

helloiamlukas:
Yes, you are right. I removed them now. Thanks for the hint!

database/migrations/create_navigations_table.php.stub
```diff
@@ -11,7 +11,8 @@ return new class extends Migration {
             $table->id();
             $table->string('handle')->unique();
             $table->string('title');
-            $table->json('settings')->nullable();
+            $table->string('locale')->nullable()->index();
```
jasonvarga, line 14: What's this `locale` column?

jasonvarga (approved):

PR #83, opened 2022-11-02
Add originBehavior to collection

ryanmitchell:
Once https://github.com/statamic/cms/pull/6983 is merged and released this will need added to eloquent.

Before merge we should also update the minimum Statamic core version in composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": "^8.0",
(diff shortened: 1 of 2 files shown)


jasonvarga:
I've updated the code and composer requirement. Once I tag cms I'll re-run the tests here and it should pass.

ryanmitchell:
Thanks!

Issue #78, opened 2022-10-20
Taxonomies doesn't support multi sites ( multi language )


Crasheng:
Creating a taxonomy that support both language doesn't work.

the dashboard enables for both language ( but one of them which is not the origin doesn't save if title filed was supporting multi lang )
 
![Screen Shot 2022-10-20 at 9 29 15 AM](https://user-images.githubusercontent.com/4809602/196872710-9d8daf10-cec5-40cc-8cc7-58223cbc4444.png)
![Screen Shot 2022-10-20 at 9 29 38 AM](https://user-images.githubusercontent.com/4809602/196872773-d4a4cef1-0ac7-472c-9003-4a16f04d0dfd.png)

ryanmitchell:
I'm not a bit multi-site user so its not entirely clear to me that the issue is? Are you saying that in your example there should also be a term for `arabic` alongside the terms for `default` ?

Crasheng:
> I'm not a big multi-site user so its not entirely clear to me that the issue is? Are you saying that in your example there should also be a term for `arabic` alongside the terms for `default` ?

yes either having an origin column like "entries"   or like "Globals" where it's being saved in the localization column (JSON), check image for reference 
![Screen Shot 2022-10-20 at 9 44 14 AM](https://user-images.githubusercontent.com/4809602/196875337-a08fe4f4-af22-4983-9ebe-e126b60b3c12.png)

But some how much of modules support the multi-site expect the taxonomies

ryanmitchell:
It does save localisations, you can see that here:
https://github.com/statamic/eloquent-driver/blob/f5fdc8175e057b3290b1fc792b0a3b56437e3aa7/src/Taxonomies/Term.php#L25-L29

But I think the issue is with line 68 here:
https://github.com/statamic/eloquent-driver/blob/f5fdc8175e057b3290b1fc792b0a3b56437e3aa7/src/Taxonomies/Term.php#L67-L71

It should be assigning to `$data['localizations'][$locale]` - do you want to try changing that on your side to see if it helps?

Crasheng:
![Screen Shot 2022-10-20 at 12 39 06 PM](https://user-images.githubusercontent.com/4809602/196913856-004eedcc-039d-44ec-96db-662770745f9c.png)

As you can see i have added `$data['localizations'][$locale] ` but still
[...]

PR #86, opened 2022-11-02
Fixed terms localization not working

FrittenKeeZ:
This will ensure localizations are stored correctly (fixes https://github.com/statamic/eloquent-driver/issues/78)


jasonvarga:
First time I've seen what the diff get this so wrong.

PR #87, opened 2022-11-10
Add composer patch instructions

edalzell:
There's an issue w/ Postgres DBs that requires a patch, until PR is merged.


jasonvarga:
We'll just merge the PR.

PR #95, opened 2022-11-22
When a collection is orderable, use that order

ryanmitchell:
As reported [over on Discord](https://discord.com/channels/489818810157891584/996788399371468861/1044251904500305930) the orderable collection view doesn't default to `order` sorting. 

So by default we need to save the `sort_field` as `order` in the database for these collections.


jasonvarga:
I think I noticed it was because `title` was being saved to the settings for `sort_field`. If nothing was saved, it would have worked fine. I think it happened during import.

The logic should be, if there's a sort field, use that.

https://github.com/statamic/cms/blob/5af6b5fa9c6b38cc8a30f595a1b510fb9245d13b/src/Entries/Collection.php#L143-L151

jasonvarga:
Ok not the importer then, but somewhere. 😄

ryanmitchell:
Should we just not be saving sortField?

jasonvarga:
Maybe we should only save it if it's "custom". ie not `order` for orderable collections, not `date` for date collections, or `title`. (Following the logic of the line I linked to before)

On the file driver, it doesn't get saved - because there's not even a field for it on the collection config screen. The only way to customize it is to put it in the yaml.

ryanmitchell:
Oh maybe it’s just my error in saving it in the first place. I think we should just not save it and let the default behaviour happen.

Issue #97, opened 2022-11-29
[Planetscale Compatibility] Add option to skip foreign key constraints 


jackmcdade:
I'm working on testing Statamic with PlanetScale and ran into this issue here:

```
 SQLSTATE[HY000]: General error: 1105 foreign key constraints are not allowed, see https://vitess.io/blog/2021-06-15-online-ddl-why-no-fk/ (SQL: alter table `form_submissions` add constraint `form_submissions_form_id_foreign` foreign key (`form_id`) references `forms` (`id`) on delete cascade)
 ```

I'm not quite sure what the best way to check for configure excluding [this line](https://github.com/statamic/eloquent-driver/blob/master/database/migrations/create_form_submissions_table.php.stub#L12). 

Thoughts?

ryanmitchell:
Id be happy to drop the foreign key entirely for what it’s worth.

jackmcdade:
Pretty low risk, right? You could always add your own constraint later if you want those cascading deletes.

ryanmitchell:
Yep

jackmcdade:
Then let's burn it with fire! 🔥

ryanmitchell:
Sure :) Here ya go: https://github.com/statamic/eloquent-driver/pull/98

ryanmitchell:
That ones maybe a bit more controversial, but I've updated the PR and Jason can make the call on it :)

FrittenKeeZ:
@jackmcdade if you delete a form, are submissions automatically deleted by Statamic as well?

jasonvarga:
> if you delete a form, are submissions automatically deleted by Statamic as well?

Yes

PR #98, opened 2022-11-29
Remove foreign id from form and entry migrations

ryanmitchell:
Fixes: #97


FrittenKeeZ (changes requested):
Removing the cascading from the submissions, we'll at least need an observer to remove any related submissions on delete, otherwise you'll end up with a bunch of orphaned submissions.

ryanmitchell:
@FrittenKeeZ Why don’t open a PR with the changes you want and ill close this one.

FrittenKeeZ:
@ryanmitchell if we just skip the orphan handling of submissions, it's a few tweaks only by slapping `->index()` on the fields in question and preserving the primary id thing for uuid entries

PR #100, opened 2022-11-30
Replaced foreign keys with simple indexes

### LINKED ISSUE #97: [Planetscale Compatibility] Add option to skip foreign key constraints

FrittenKeeZ:
Also added index to entries order, and re-aligned migrations with Laravel's stub.

Fixes #97

### DIFF
--- a/database/migrations/create_entries_table_with_string_ids.php.stub
+++ b/database/migrations/create_entries_table_with_string_ids.php.stub
@@ -4,31 +4,38 @@ use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 use Statamic\Eloquent\Database\BaseMigration as Migration;
 
-return new class extends Migration {
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
     public function up()
     {
         Schema::create($this->prefix('entries'), function (Blueprint $table) {
             $table->uuid('id');
             $table->string('site')->index();
-            $table->uuid('origin_id')->nullable();
+            $table->unsignedBigInteger('origin_id')->nullable()->index();
             $table->boolean('published')->default(true);
             $table->string('status');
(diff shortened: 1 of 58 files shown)


database/migrations/create_entries_table_with_string_ids.php.stub
```diff
@@ -10,22 +10,18 @@ return new class extends Migration {
         Schema::create($this->prefix('entries'), function (Blueprint $table) {
             $table->uuid('id');
             $table->string('site')->index();
-            $table->uuid('origin_id')->nullable();
```
jasonvarga, line 13: Was thing change from `uuid` to `unsignedBigInteger` intentional? The origin id would also be a uuid.

jasonvarga:
Sorry I deleted my comment because right after I realized how long this has been opened, I didn't want to bother you. I was going to just sort it out myself. 😊 But thanks for updating that.

ryanmitchell:
@jasonvarga I know you've got a seperate fix for the FK errors at https://github.com/statamic/eloquent-driver/pull/160 but I still think this is worth merging as it solves Jack's original issue. Do you have any concerns?

ryanmitchell:
@FrittenKeeZ I've added an update script to provide migrations for existing installs so that the table definitions match going forward.

Issue #96, opened 2022-11-24
Model attribute not retrieved


Z3d0X:
When using 
```php
Model::preventAccessingMissingAttributes(! $this->app->isProduction())
```

Im getting an exception:
```
The attribute [origin_id] either does not exist or was not retrieved for model [Statamic\Eloquent\Entries\UuidEntryModel].
```

Flare Exception Url: https://flareapp.io/share/dmk4YYDP#F98

I understand this was added recently https://github.com/laravel/framework/pull/44283

PR #102, opened 2022-12-22
Fix: model atrribute not retrieved error

Z3d0X:
fixes https://github.com/statamic/eloquent-driver/issues/96


src/Entries/EntryQueryBuilder.php
```diff
@@ -62,6 +62,10 @@ public function find($id, $columns = ['*'])
 
     public function get($columns = ['*'])
     {
+        if ($columns !== ['*']) {
+            $columns = array_unique(array_merge($columns, ['site', 'origin_id', 'slug', 'collection', 'created_at', 'updated_at']));
```
ryanmitchell, line 66: It would be better to merge with `self::COLUMNS` to avoid repetition? But really with this change youre just getting all the columns anyway.

Issue #112, opened 2023-01-30
Querying missing localization data doesn't work


jasonvarga:
On localized entries, any fields that you don't choose to localize will fall back to what's in the origin. That's fine.

However if you try to perform a query on one of these fields, the localized versions will not appear.

For example: (using yaml to represent the rows, obviously in eloquent it's stored differently)

```yaml
id: 1
title: Origin entry
foo: bar
```

```yaml
id: 2
origin_id: 1
title: Localized entry
# foo is intentionally missing, because we want it to fall back to bar
```

```php
Entry::query()->where('foo', 'bar')->get(); // only returns the origin entry
```

In the Stache driver it works fine because the indexes/columns are generated on the fly.

An idea: store _all_ the data so it's queryable, but also store something like `localized_keys` so we can work out what goes in `$entry->data()` when we build the objects.

ryanmitchell:
That would work but it would mean we also need to re-save any localisations when the origin is saved so the new values propagate.

The only other option I can think of is updating the query builder to also look for values on the parent (using `->whereHas()` or something similar). But that would get messy.

PR #114, opened 2023-01-30
Store origin data alongside localized data

ryanmitchell:
As discussed [here](https://github.com/statamic/eloquent-driver/issues/112), we should store origin entry data alongside localisation data so it can be queried at the database level.

Let me know if youre happy with this approach as I'll need to add some test coverage on this before it should be merged.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -13,7 +13,7 @@ class Entry extends FileEntry
 
     public static function fromModel(Model $model)
     {
-        $data = isset($model->data['localized_fields']) ? collect($model->data)->only($model->data['localized_fields']) : $model->data;
+        $data = isset($model->data['__localized_fields']) ? collect($model->data)->only($model->data['__localized_fields']) : $model->data;
 
         $entry = (new static())
             ->origin($model->origin_id)
@@ -69,7 +69,7 @@ class Entry extends FileEntry
 
                 $data = $origin->data()->merge($data);
 
-                $data->put('localized_fields', $localizedFields);
+                $data->put('__localized_fields', $localizedFields);
 
                 if (! in_array('date', $localizedFields)) {
                     $date = $origin->hasDate() ? $origin->date() : null;
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -81,6 +81,10 @@ class Entry extends FileEntry
             }
         }
 
+        if ($parent = $this->parent()) {
+            $data->put('parent', (string) $parent->id);
+        }
+
         $attributes = [
             'origin_id'  => $origin?->id(),
             'site'       => $this->locale(),
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -49,13 +49,23 @@ class Entry extends FileEntry
             }
 
             if ($origin) {
-                $localizedFields = $blueprint
+                $localizedBlueprintFields = $blueprint
                     ->fields()
                     ->localizable()
                     ->all()
                     ->map
                     ->handle()
                     ->all();
+
+                // remove any fields in entry data that are marked as localized but value is blank
+                $localizedFields = [];
+                foreach ($localizedBlueprintFields as $blueprintField) {
+                    if ($data->get($blueprintField) == '') {
+                        $data->forget($blueprintField);
+                    } else {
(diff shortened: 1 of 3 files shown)


ryanmitchell:
@FrittenKeeZ yep, I was battling the tests a bit so moved this to draft. Should be working now with the latest commit.

src/Entries/Entry.php
```diff
@@ -36,17 +38,39 @@ public function toModel()
...
+            }
+
+            if ($origin) {
+                $localizedFields = $blueprint
+                    ->fields()
+                    ->localizable()
+                    ->all()
+                    ->map
+                    ->handle()
+                    ->all();
```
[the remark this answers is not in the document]
ryanmitchell: So get the data for the entry for those fields and filter it to remove blank values?

FrittenKeeZ:
@ryanmitchell @jasonvarga we're facing rather big performance issues with `blueprint` being stored in json instead of its own field, so I just wanted to hear you out if you're okay with pulling `blueprint` out of json with an index so queries don't take 5+ seconds to complete?

src/Entries/Entry.php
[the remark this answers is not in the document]
ryanmitchell: I've gone ahead and [implemented this](https://github.com/statamic/eloquent-driver/pull/114/commits/352427fd2ef8a6a9ec73957327333b34eb55fdae) - let me know if you want a different approach

FrittenKeeZ:
@ryanmitchell do you have any sense of whether this is ready for use? The missing `blueprint` on translated entries are causing us big issues currently, so I'm torn between using this patch to get the missing data or creating a PR for storing `blueprint` in a separate field in the database for improving performance as well.

ryanmitchell:
As far as I can see it is, but Jason may feel differently

FrittenKeeZ:
We just talked it through with the team and I'll be making a PR to fix #126 soon, as it will fix three issues we have regarding performance, code complexity and data duplication for blueprint.

src/Entries/Entry.php
```diff
@@ -42,15 +44,48 @@ public function toModel()
 
         $data = $this->data();
 
+        $origin = $this->origin();
+        $date = $this->hasDate() ? $this->date() : null;
+
+        if ($blueprint = $this->blueprint()) {
```
FrittenKeeZ, line 50: @ryanmitchell We ran into a few issues with this block regarding fields not properly being detected as inherited, which is why I made a few changes to my PR.
https://github.com/statamic/eloquent-driver/pull/144/files#diff-31fcf8a95fd0dbc8da9e8dc7ae856aa51bd921c50a8059e23df7d411737ab65fR47

ryanmitchell: Can you explain the issues you were seeing?

FrittenKeeZ: Reverting through CP - I guess through code it would just be unsetting the fields?

ryanmitchell: Got you now, and was able to replicate it.
I've pushed a fix for that.

FrittenKeeZ:
@ryanmitchell I can see I added some additional handling regarding `parent`, but I can't remember whether it was an issue or I just added it as a precaution.
https://github.com/statamic/eloquent-driver/pull/144/files#diff-31fcf8a95fd0dbc8da9e8dc7ae856aa51bd921c50a8059e23df7d411737ab65fR71

ryanmitchell:
Yeah we're not storing parent going forward - its redundant as we have origin as a column.

FrittenKeeZ:
> Ah right, so german might be the parent of french, but english is the fallback? Ok, storing the ID is fine (as your PR did) - I think until this PR its storing the whole entry which is a lot.

Not quite... each site / language is its own regarding parents, like so:
```
                 Origin
Contact (EN)  - - - - - - - Kontakt (DE)
    |                         |
    | Parent                  | Parent
    |            Origin       |
About US (EN) - - - - - - - Über Uns (DE)
```

ryanmitchell:
@FrittenKeeZ are you happy that this covers everything you found now?

FrittenKeeZ (approved):

ryanmitchell:
Thanks @FrittenKeeZ

PR #129, opened 2023-02-28
Fix entries blueprint performance

### LINKED ISSUE #126: Performance issues with blueprint as JSON field
@ryanmitchell @jasonvarga we're facing rather big performance issues with `blueprint` being stored in json instead of its own field, so I just wanted to hear you out if you're okay with pulling `blueprint` out of json with an index so queries don't take 5+ seconds to complete?

_Originally posted by @FrittenKeeZ in https://github.com/statamic/eloquent-driver/issues/114#issuecomment-1420531869_

FrittenKeeZ:
Fixes #126

### DIFF
--- a/database/migrations/create_entries_table.php.stub
+++ b/database/migrations/create_entries_table.php.stub
@@ -18,6 +18,7 @@ return new class extends Migration {
             $table->string('date')->nullable();
             $table->integer('order')->nullable();
             $table->string('collection')->index();
+            $table->string('blueprint', 30)->nullable()->index();
             $table->jsonb('data');
             $table->timestamps();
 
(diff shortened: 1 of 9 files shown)


FrittenKeeZ:
This needs to be re-run when `3.4.5` release is unbotched

maxi032:
When are you guys going to merge this as we are also facing the same problems

ryanmitchell:
@FrittenKeeZ I've just applied this to a 2.1.0 install and I would be worried people wont catch that they need to run the `php please eloquent:migrate-entries-blueprint` command unless they are keeping an eye on merged PRs.

Why don't we run it as part of the migration?

ryanmitchell:
Great, leave it with me and I'll update

PR #91, opened 2022-11-15
Fixed Variables origin not being serialized as string

FrittenKeeZ:
Ensure that Variables origin is properly serialized to locale string instead of a Variables object

PR #130, opened 2023-03-03
Cast global set origin attribute to a collection

royduin:
As it tries to get the values from it with `values()` but it's an array so you'll get:
```
Call to a member function values() on array
```

See: https://github.com/statamic/cms/blob/8d5b25d7c264bbacfebda2a1ff6ae412b4bb0b02/src/Data/HasOrigin.php#L13

### DIFF
--- a/src/Globals/VariablesModel.php
+++ b/src/Globals/VariablesModel.php
@@ -11,7 +11,9 @@ class VariablesModel extends BaseModel
 
     protected $table = 'global_set_variables';
 
-    protected $casts = [];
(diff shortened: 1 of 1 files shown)


jasonvarga:
This stops the error but the origin is still incorrect and doesn't work. I'm working on a fix. Thanks!

royduin:
How can I keep track of this? Should I create an issue for it?

jasonvarga:
You can keep track of #91 which was fixing the same thing.

PR #133, opened 2023-03-09
Split configs for asset containers, collection trees and navigation trees

piljac1:
Fixes https://github.com/statamic/eloquent-driver/issues/147

This small pull request enables the possibility to define a different driver for collections and collection trees.

This can be useful if you want your collection definitions to be stored as flat files (so they're available and synced throughout all environments), but have eloquent entries which has a direct impact on the tree structure (which would not be the same in all environments).

All tests are passing, but I saw no config-based tests I could alter or enrich. However, it was tested in all 5 possible configurations on a local site:
- driver -> eloquent
tree_driver -> commented

- driver -> file
tree_driver -> commented

- driver -> eloquent
tree_driver -> eloquent

- driver -> eloquent
tree_driver -> file

- driver -> file
tree_driver -> eloquent

Let me know if you need me to do some changes !

**P.S.** This is a non breaking change, so people that don't have the new key in their published config file won't be affected.

**Second P.S.** My intent is to do the same kind of changes where a split would be possible for other "models" (i.e. navigations) without introducing breaking changes.

### DIFF
--- a/config/eloquent-driver.php
+++ b/config/eloquent-driver.php
@@ -18,12 +18,14 @@ return [
     ],
 
     'collections' => [
-        'driver'       => 'eloquent',
-        'model'        => \Statamic\Eloquent\Collections\CollectionModel::class,
-        // Uncomment the setting below if you need to specify a different driver for your collections tree.
-        // 'tree_driver' => 'eloquent',
-        'tree'         => \Statamic\Eloquent\Structures\CollectionTree::class,
-        'tree_model'   => \Statamic\Eloquent\Structures\TreeModel::class,
+        'driver' => 'eloquent',
+        'model'  => \Statamic\Eloquent\Collections\CollectionModel::class,
+    ],
+
+    'collection_trees' => [
+        'driver' => 'eloquent',
+        'model'  => \Statamic\Eloquent\Structures\TreeModel::class,
+        'tree'   => \Statamic\Eloquent\Structures\CollectionTree::class,
     ],
 
     'entries' => [
--- a/config/eloquent-driver.php
+++ b/config/eloquent-driver.php
@@ -5,10 +5,14 @@ return [
     'connection'   => env('STATAMIC_ELOQUENT_CONNECTION', ''),
     'table_prefix' => env('STATAMIC_ELOQUENT_PREFIX', ''),
 
+    'asset_containers' => [
+        'driver' => 'eloquent',
+        'model'  => \Statamic\Eloquent\Assets\AssetContainerModel::class,
+    ],
+
     'assets' => [
-        'driver'          => 'eloquent',
-        'container_model' => \Statamic\Eloquent\Assets\AssetContainerModel::class,
-        'model'           => \Statamic\Eloquent\Assets\AssetModel::class,
+        'driver' => 'eloquent',
+        'model'  => \Statamic\Eloquent\Assets\AssetModel::class,
     ],
 
     'blueprints' => [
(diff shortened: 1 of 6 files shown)


config/eloquent-driver.php
```diff
@@ -18,10 +18,12 @@
...
 
     'collections' => [
-        'driver'     => 'eloquent',
-        'model'      => \Statamic\Eloquent\Collections\CollectionModel::class,
-        'tree'       => \Statamic\Eloquent\Structures\CollectionTree::class,
-        'tree_model' => \Statamic\Eloquent\Structures\TreeModel::class,
+        'driver'       => 'eloquent',
+        'model'        => \Statamic\Eloquent\Collections\CollectionModel::class,
+        // Uncomment the setting below if you need to specify a different driver for your collections tree.
+        // 'tree_driver' => 'eloquent',
```
ryanmitchell, line 24: This should be the default to ensure backwards compatibility for existing eloquent driver users. Can you update?

I also wonder if it might be 'cleaner' to have a new config for `collection_trees` rather the bundling this under collections?

piljac1:
@ryanmitchell Thanks for the feedback! I just applied the changes. Sorry for the long wait, I have tons of work right now haha.

src/ServiceProvider.php
```diff
@@ -160,15 +161,31 @@ private function registerCollections()
...
+        }
 
         Statamic::repository(CollectionTreeRepositoryContract::class, CollectionTreeRepository::class);
 
         $this->app->bind('statamic.eloquent.collections.tree', function () {
-            return config('statamic.eloquent-driver.collections.tree');
+            // Use the old related config as a fallback to ensure backwards compatibility.
+            $collectionsOldTree = config('statamic.eloquent-driver.collections.tree');
+
+            return config('statamic.eloquent-driver.collection_trees.tree', $collectionsOldTree);
```
ryanmitchell, line 181: You could make this a one-liner as follows:

`return config('statamic.eloquent-driver.collection_trees.tree', config('statamic.eloquent-driver.collections.tree'));`

src/ServiceProvider.php
```diff
@@ -160,15 +161,31 @@ private function registerCollections()
...
         });
 
         $this->app->bind('statamic.eloquent.collections.tree_model', function () {
-            return config('statamic.eloquent-driver.collections.tree_model');
+            // Use the old related config as a fallback to ensure backwards compatibility.
+            $collectionsOldTreeModel = config('statamic.eloquent-driver.collections.tree_model');
+
+            return config('statamic.eloquent-driver.collection_trees.model', $collectionsOldTreeModel);
```
ryanmitchell, line 188: Same one-liner approach as suggested before

ryanmitchell (changes requested):

ryanmitchell:
@piljac1 sorry for the delay in looking at this.

I've pushed a change in approach. It checks that the tree_model config exists, which if it does shows that we're using the legacy keys, and so should defer any config calls to those. With the previous approach it was deferring to the collection config which when using the new config may not be what you want.

I've also gone ahead and added support for splitting navs and trees which makes sense as part of the same change.

Let me know what you think.

ryanmitchell (commented):
See my refactor.

piljac1:
Sounds good! I also aimed at splitting navs and other things, but I attempted to split navs briefly last week (applied the same logic that was done with collections) and it didn't seem to work properly (exceptions were encountered). Have you tested the nav changes you've implemented in a Statamic project (note that I tested my fix attempt on a Statamic v3.3 project and not a v4+, so it might work properly on v4+)?

ryanmitchell:
I've pushed some fixes for the nav situation

ryanmitchell:
Just leaving a note here so its not missed, before this merges we will need to update the imports for assets, collections and navs to allow users to decide what they want to import.

PR #134, opened 2023-03-10
Add presets to asset container settings

ajcsilva:
This PR adds the ability to save image presets in the Asset Container settings.


jasonvarga (approved):
I made a change so that it uses the properties instead of methods within `toModel()`.

Otherwise, it was saving some incorrect settings.

PR #136, opened 2023-03-15
Change blink handle for 'all' as it clashes with handles of 'all'

### LINKED ISSUE #135: Assets can not be stached after being moved to database
I am migrating our site's content from flat files to a Postgres database to improve performance. The site works and the files are migrated, but every time I try to `run php please stache:refresh`, I get the following error when the `$handle = 'all'`:

```
 TypeError

  Statamic\Eloquent\Assets\AssetContainerRepository::findByHandle(): Return value must be of type ?Statamic\Contracts\Assets\AssetContainer, Illuminate\Support\Collection returned

  at vendor/statamic/eloquent-driver/src/Assets/AssetContainerRepository.php:41
     37▕                 return null;
     38▕             }
     39▕
     40▕             return app(AssetContainerContract::class)->fromModel($model);
  ➜  41▕         });
     42▕     }
     43▕
     44▕     public function make(string $handle = null): AssetContainerContract
     45▕     {

      +33 vendor frames
  34  please:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```
When I access the site without the stache is it able to get the assets using the individual $handles, and when I hardcode the $handle inside `AssetContainerRepository` it works fine but it fails on `$handle = 'all'`. I presume it is because it gets a collection of AssetContainers but is there a way to fix this?

I also get the following when trying to access assets in the control panel:

```
TypeError
Statamic\Eloquent\Assets\AssetContainerRepository::all(): Return value must be of type Illuminate\Support\Collection, Statamic\Eloquent\Assets\AssetContainer returned (View: /Users/martin/Sites/sd-statamic/vendor/statamic/cms/resources/views/layout.blade.php) 
```

**Setup:**

Statamic 3.4.5 Pro
Laravel 8.83.27
PHP 8.1.16
Stache Watcher Enabled
Static Caching Disabled
adapt/tfa 1.0.1
statamic/eloquent-driver 1.2.0
thehome/statamic-elasticsearch dev-ADP-1213-loadtide-statamic-update

ryanmitchell:
Fixes #135

### DIFF
--- a/src/Assets/AssetContainerRepository.php
+++ b/src/Assets/AssetContainerRepository.php
@@ -25,7 +25,7 @@ class AssetContainerRepository extends StacheRepository
 
     public function findByHandle(string $handle): ?AssetContainerContract
     {
-        return Blink::once("eloquent-assetcontainers-{$handle}", function () use ($handle) {
+        return Blink::once("eloquent-assetcontainers-handle-{$handle}", function () use ($handle) {
             $model = app('statamic.eloquent.assets.container_model')::whereHandle($handle)->first();
 
             if (! $model) {
--- a/src/Assets/AssetContainerRepository.php
+++ b/src/Assets/AssetContainerRepository.php
@@ -16,7 +16,7 @@ class AssetContainerRepository extends StacheRepository
         return Blink::once('eloquent-assetcontainers', function () {
             return app('statamic.eloquent.assets.container_model')::all()
                 ->map(function ($model) {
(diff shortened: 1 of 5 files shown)


jasonvarga:
Sorry for the delay.

Looks like the conflict only happened for asset containers because the blink key for singular containers used the plural.

e.g. `assetcontainers-something`

Everywhere else, for the singular, it used the singular.

e.g. `collection-something`

What you've done is still a good change to make though. 👍

PR #138, opened 2023-03-27
Only save custom sort field and direction

ryanmitchell:
@robdekort found a lovely bug that when you have a dated collection and make it order able it wasn't honouring the order you set. So this fix changes it by not saving the `sortField` to the database when the collection is orderable.

### DIFF
--- a/src/Collections/Collection.php
+++ b/src/Collections/Collection.php
@@ -68,7 +68,7 @@ class Collection extends FileEntry
                 'default_status'       => $source->defaultPublishState,
                 'structure'            => $source->structureContents(),
                 'sort_dir'             => $source->sortDirection(),
-                'sort_field'           => $source->sortField(),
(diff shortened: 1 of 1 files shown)


jasonvarga:
We should only save sortField (and sortDirection) if it's a **custom** value.

By leaving it blank, it'll apply the appropriate logic.

The custom value would be in the property. We could probably add a method to the Collection class like `customSortField` and then change this line to `'sort_field' => $source->customSortField()`.

jasonvarga:
In order for this fix to be reflected, you'll need get rid of the custom sort field and direction.

You can run `php artisan tinker` and run this for each collection, substituting the collection handle appropriately.

```php
\Statamic\Eloquent\Collections\CollectionModel::where('handle', 'your_collection')->update([
  'settings->sort_field' => null,
  'settings->sort_dir' => null,
]);
```

Or, put it in a migration.

PR #142, opened 2023-04-11
v4 support

ryanmitchell:
A PR to handle changes required to support v4 of core

[x] Support tabs at top level in blueprints with sections beneath

### DIFF
--- a/src/Commands/ImportBlueprints.php
+++ b/src/Commands/ImportBlueprints.php
@@ -72,13 +72,22 @@ class ImportBlueprints extends Command
 
             $contents = YAML::file($path)->parse();
             // Ensure sections are ordered correctly.
-            if (isset($contents['sections']) && is_array($contents['sections'])) {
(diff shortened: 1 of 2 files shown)


src/Entries/Entry.php
```diff
@@ -17,8 +17,8 @@ public static function fromModel(Model $model)
             ->origin($model->origin_id)
             ->locale($model->site)
             ->slug($model->slug)
-            ->date($model->date)
             ->collection($model->collection)
+            ->date($model->date)
```
jasonvarga, line 21: You'll wanna move this a bit lower and wrap in a condition.

e.g. 

```
if ($model->date) {
  $entry->date($model->date);
}
```

If you try to set `date()` on a non-dated collection entry, you'll get the error mentioned in https://github.com/statamic/cms/issues/7959

I'm not 100% sure, but I think `$model->date` would only ever exist if the collection is dated, so just checking if the attribute isn't null is enough.

PR #145, opened 2023-04-14
Fixing some model references to configurable references

dimitri-koenig:
There are some references to models which are not called via app('...') but directly. That means that overriding those references is not possible. This PR fixes that.


jasonvarga:
Thank you

PR #156, opened 2023-05-24
Fix to hyphenated taxonomies from uri

Alt-Ben:
Super tiny change to fix the hyphenated taxonomies in term uris.


jasonvarga:
Can you explain the situation this is fixing? This isn't in the core so it might also be happening there.

ryanmitchell:
Ok for now Ive opted for (1) as its the path of least friction for now.
Let me know if this works for your use case.

ryanmitchell:
@Alt-Ben did you get a chance to review this fix?

Alt-Ben:
Sorry! Just got around to checking - all looks good from my side! Thank you!

PR #158, opened 2023-06-02
Update entry order when saving a tree and collections are stored in files

### LINKED ISSUE #139: An orderable dated collection can't be custom sorted
When you have a dated orderable collection you can't custom sort it. Well you can in the UI but the order won't persist. 

Was talking about this on Discord and it turns out @ryanmitchell already made a PR.

jasonvarga:
Fixes #139

This adds an event listener that does the exact same thing as the core's UpdateStructuredEntryOrder listener.

It'll update the `order` column of entries when you save a tree.

It will only do it if entries are configured to be in Eloquent, and Collections are configured to not be.

### DIFF
--- a/src/Listeners/UpdateStructuredEntryOrder.php
+++ b/src/Listeners/UpdateStructuredEntryOrder.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Statamic\Eloquent\Listeners;
+
+use Statamic\Eloquent\Collections\CollectionRepository;
+use Statamic\Events\CollectionTreeSaved;
+
+class UpdateStructuredEntryOrder
+{
+    public function handle(CollectionTreeSaved $event)
+    {
+        $tree = $event->tree;
+        $collection = $tree->collection();
+
+        if (config('statamic.eloquent-driver.entries.driver', 'file') !== 'eloquent') {
+            return;
+        }
+
+        if (config('statamic.eloquent-driver.collections.driver') === 'eloquent') {
+            // If the collections are configured to use Eloquent, then the entry
+            // order will be updated through the regular event/listener flow.
+            return;
+        }
+
+        $diff = $tree->diff();
+
+        $ids = array_merge($diff->moved(), $diff->added());
+
+        if (empty($ids)) {
+            return;
+        }
+
+        app(CollectionRepository::class)->updateEntryOrder($collection, $ids);
+    }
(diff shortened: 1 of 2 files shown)


modrictin:
Hey @jasonvarga ,

what happened before you introduced this change?

I'm thinking of removing it for my use case. I tried to update to latest version and I have some issues with this update.

More on the issue here, in my discord message:

Hey everyone,

I'm currently facing an issue while upgrading from Eloquent driver v1.2 to v2.3.1 and could really use some assistance.

The problem stems from the introduction of an order column in the database. Whenever I save an entry, the UpdateStructuredEntryOrder listener gets triggered because I'm exclusively using the Eloquent driver for managing the entries. As a result, the Eloquent CollectionRepository tries to update the order column for all entries in that collection.

Now, here's the challenge—I have a large number of entries, with over 110,000 in my database, and a few thousand in a single collection. Unfortunately, using the queue driver sync to save the entry isn't a viable solution as it tries to process an enormous amount of entries, causing significant delays. While I'm using a different queue driver in my production environment, I still want to avoid this situation altogether.

I'm looking for any advice, ideas, or insights on how to handle this issue effectively. One thought I had was to remove the logic responsible for updating the order column for the saved entries.

Before proceeding, I have a few questions:

Is it true that the order column is exclusively used for managing the order of entries within a collection in the EntriesController::index() method?
Are there any other components or processes in my system that rely on the order column?

If I remove the logic that updates the order column, will it impact any other functionality or disrupt the expected behavior of my application?

Thank you in advance for your help!
How did this functionality even work before introducing this order column. If it did not, that is ok with me, I cold just remove it then


Thank you so much!
Tin

ryanmitchell:
The order is only used in a manually ordered collection - I suspect if you have that many entries you’re not ordering it manually?

the event should not run if you turn off the sortable config setting for the collection.

modrictin:
Hey @ryanmitchell that's what I thought!

Thank you so much for the confirmation 😊

ryanmitchell:
If you are finding it’s running on a not structured collection that’s definitely a big and we should look into that.

modrictin:
Hey @ryanmitchell ,

I did not find anything unordinary, but I'm still experiencing slow entry saving.

Apparently this is causing it:
image

Why is this introduced all of a sudden? 

Thanks,
Tin

Issue #128, opened 2023-02-22
Split globalset configuration from data (variables)


royduin:
For some functions; for example global sets the configuration is stored with the data. Are there any plans to split this? Or will a PR with this (breaking) change be accepted? We'd like to have all configuration like blueprints, global sets, etc in files and all data in the database. Currently we've set `statamic.eloquent-driver.blueprints.driver` to `file` and `statamic.eloquent-driver.global_sets.driver` to `eloquent`. This way we can version control the configuration of the blueprints but can't do this with the global set configuration. Currently this is stored like:
```
{"default": {"data": {"field": "DATA"}, "locale": "default", "origin": null}, "site2": {"data": [], "locale": "site2", "origin": "default"}}
```

ryanmitchell:
I'm not totally sure what you are wanting to achieve in splitting? Can you try and explain it a different way?

royduin:
Statamic uses flat file by default; with a lot of entries this can become slow so that's why this Eloquent driver exists. As we're working with multiple environments (DTAP) we like to version control the configuration. Otherwise when we create a field, collection, global, etc and everything is stored within the database we've to redo this on every environment. 

If we take global sets as an example; when I create a new set with some fields the blueprints are stored on the disk (as I configured it that way) but everything else is stored within the database. So when I commit and push this (so that's just the blueprint); the test environment doesn't have my new global set as it is stored within the database.

It would be nice to have all configuration in files and just the data in the database so all configuration can be version controlled.

Thoughts?

ryanmitchell:
So on Globals you want the 'global set' (handle, title etc) to be stored in file, but the 'variables' within the global to be in the database?

royduin:
Correct, but this also applies to the navigation and probably more parts.

ryanmitchell:
I'd have no objection to that as long as it's configurable and you're happy to PR it.

It wouldnt be a breaking change if you make it an option in for each repository in the config file (which defaults to the current behaviour).

royduin:
Aweomse! Am I the first one reporting something like this? Curious how others work with this. Maybe we've to change our workflow? I mean; if everything is stored within the database you've to redo your changes on all environments where you end up merging data as the end user / customer continues to edit data on production.

ryanmitchell:
First time I've seen it be asked. We tend to seed data so this doesn't affect our workflows.

royduin:
[...]

PR #168, opened 2023-06-15
Restructure global variables to use its own table

ryanmitchell:
Following on from the conversation in https://github.com/statamic/eloquent-driver/issues/128, it makes sense to split globals and variables into separate tables, enabling splitting of configuration from data.

Ideally we would be able to offer a config for letting you opt into eloquent globals and variables separately, but from what I can see this is not possible in core as things stand. Scope for future development.

Closes: https://github.com/statamic/eloquent-driver/issues/128

### DIFF
--- a/database/migrations/create_global_variables_table.php.stub
+++ b/database/migrations/create_global_variables_table.php.stub
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Statamic\Eloquent\Database\BaseMigration as Migration;
+
+return new class extends Migration {
+    public function up()
+    {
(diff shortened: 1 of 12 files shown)


royduin:
Looking good to me! Curious what the challenge would be to separate it?

ryanmitchell:
Not a lot to be honest, working on a PR to core at the moment

ryanmitchell:
@FrittenKeeZ I'd love a second pair of eyes on this if you had the time to check it over?

FrittenKeeZ (approved):
Looks good - I don't have an option to test it out currently though

ryanmitchell:
@royduin https://github.com/statamic/cms/pull/8343 if you want to follow along

royduin:
Subscribed 👌🏻

Issue #26, opened 2021-08-09
Content API: Selecting Fields not working / error


helloiamlukas:
## Bug Description
If you try to select just specific fields (exactly as described [in the documentation](https://statamic.dev/content-api#selecting-fields)), nothing happens or an error will be thrown.

#### Request
`http://statamic.test/api/collections/newsletters/entries?fields=title`
#### Response
<img width="854" alt="Bildschirmfoto 2021-08-09 um 14 00 31" src="https://user-images.githubusercontent.com/15997450/128702931-d87d3f4c-2760-49ee-986f-133493367ae0.png">
I would expect only the `title` attribute of each entry shown here.

#### Request
`http://statamic.test/api/collections/newsletters/entries?fields=id`
or
`http://statamic.test/api/collections/newsletters/entries?fields=date`
or
`http://statamic.test/api/collections/newsletters/entries?fields=uri`
#### Response
**Error: Call to a member function cascade() on null**
<img width="1222" alt="Bildschirmfoto 2021-08-09 um 14 09 34" src="https://user-images.githubusercontent.com/15997450/128703857-2be24ab8-a7fd-4ddb-851c-959b187dc6dd.png">

## Environment

Statamic 3.1.32 Pro
Laravel 8.40.0
PHP 7.4.19
statamic/collaboration 0.1.0

**Install method (choose one):**
- Fresh install from `statamic/statamic`

Smef:
This is something we're experiencing as well. Including the "collection" field seems to be a workaround for now, but a real fix would be great!

ryanmitchell:
I think the solution will be to always select collection even when it's not asked for. But yep leave it with me and I'll try and come up with something before the next release.

ryanmitchell:
I've opened a [pull request here](https://github.com/statamic/eloquent-driver/pull/183) which should solve the issue, though I need to spend some time sitting on it to decide if (a) there isnt a better way, (b) this doesnt affect other areas of the driver

Smef:
It turns out adding the collection field doesn't actually resolve this. The query doesn't fail, but all the fields of the entry are null when the collection field is included.

ryanmitchell:
@Smef did you try applying the changes made here: https://github.com/statamic/eloquent-driver/pull/183

Smef:
I tried updating the EntryQueryBuilder.php file with the changes in that PR. It didn't seem to make a difference with the returned data.
Here's an example
`fields=title,price_base`
returns:
```
{
[...]

PR #183, opened 2023-06-29
Fix issues with selecting fields in the content API

ryanmitchell:
I don't love this approach, but it seems like it might be the only thing that works. 
Switch to always getting all data from the database, then limiting whats returned by selectedQueryFields in the Entry class.

Fixes https://github.com/statamic/eloquent-driver/issues/26

### DIFF
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -80,7 +80,7 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
     {
         $items = EntryCollection::make($items)->map(function ($model) use ($columns) {
             return app('statamic.eloquent.entries.entry')::fromModel($model)
-                ->selectedQueryColumns($columns);
+                ->selectedQueryColumns($this->selectedQueryColumns ?? $columns);
         });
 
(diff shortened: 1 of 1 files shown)


ryanmitchell:
@jasonvarga could you sanity check this when you have some time. I really don’t like getting all the data every query but I feel like there may be no other option.

jasonvarga:
> can you see a reason why we shouldn't just get the full row each time

Performance, I guess. The same reason you don't just use `select *` all the time.

As for #96, we could add `origin_id` along with `collection`. Clearly `collection` isn't the only required field.

jasonvarga:
Sorry to be clear, initially I suggested "enforce the `collection` column is selected". What I really mean is that we need "enforce a list of columns".

If you do `$query->get('title')`, we should merge on `collection`, `origin_id`, maybe `slug`, etc.

jasonvarga:
Er wait I see what you're saying. The rest of the custom fields are all stored inside `data`. I forgot it worked that way! The important stuff is all dedicated columns. Ok I guess we just get them all 🤷

ryanmitchell:
Glad it has you going round in circles too! I cant decide whats best either.

PR #190, opened 2023-07-21
Fix fieldsets export

clementmas:
I tried to export my blueprints and fieldsets back to the file system but there was an error. So my blueprint were exported but not the fieldsets.

```
Statamic\Fields\FieldsetRepository::find(): Argument #1 ($handle) must be of type string, null given
```

This change fixed the issue and allowed me to export everything properly.

Sets to reproduce:
- php please eloquent:import-blueprints
- php please eloquent:export-blueprints


ryanmitchell:
Thank you!

PR #178, opened 2023-06-24
Split globals and global variables, provide eloquent global variable repository

ryanmitchell:
If/when https://github.com/statamic/cms/pull/8343 lands in core we need to be ready to support an eloquent global variables repository.  

This draft PR is a work in progress to move towards it. 
It also relies on https://github.com/statamic/eloquent-driver/pull/168 being merged.

Closes https://github.com/statamic/eloquent-driver/issues/172

PR #193, opened 2023-07-31
Don't double-save descendants

jasonvarga:
In statamic/cms#8505 descendants will be saved. There is no longer a need to do it from the Eloquent repository side. It would only cause them to be saved twice.

I'm bumping the requirement of cms to the version where that PR would be available.

This is not currently an issue, but will be once that PR is released. It's only minor - the entries would be double saved, and EntrySaved events dispatched multiple times. Database `update` queries wouldn't even run since the model would have no changes the second time.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require": {
         "php": "^8.0",
-        "statamic/cms": "^4.14"
+        "statamic/cms": "^4.16"
     },
     "require-dev": {
         "doctrine/dbal": "^3.3",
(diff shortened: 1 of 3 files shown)


jasonvarga:
This PR will fail until #178 is completed.

ryanmitchell:
Seems like a couple of the entry tests are failing now as the data isn't propagating to the database?
`it_propagates_origin_date_to_descendent_models` and `it_propagates_updating_origin_data_to_descendent_models`

jasonvarga:
Looks like it was because the test was saving through the repository rather than the entry. The descendant-saving logic is happening on the entry now. It seems like the the reason the repository was maybe just a copy paste error from somewhere else, because the entry is saved normally a few lines earlier.

Issue #198, opened 2023-08-25
orderable config throws an expception at CollectionTreeRepository


aknEvrnky:
Hi, I think I've found a bug that causes critic error. To reproduce it:

- create a new collection.
- set Orderable to true.
- try to reorder, create or any entry transaction on created collection.
- Call to undefined method Statamic\\Structures\\CollectionTree::toModel() at vendor/statamic/eloquent-driver/src/Structures/CollectionTreeRepository.php:25

my eloquent driver config:

| model      | driver |
| ----------- | ----------- |
| asset_containers      | eloquent       |
| assets   | eloquent        |
| blueprints   | file        |
| collections   | file        |
| collection_trees   | eloquent        |
| entries   | eloquent        |

rest of drivers are eloquent.

## my dependency & environment setup
```
php: 8.2
laravel: 10.20
statamic/cms: 4.19
statamic/eloquent-driver: 2.6.1
```

### the temporary solution
I've changed my collection_trees driver to file.

ryanmitchell:
@aknEvrnky I've put a fix in [place here,](https://github.com/statamic/eloquent-driver/pull/199) do you want to check it out?

ryanmitchell:
Should be fixed in 2.7.0

PR #199, opened 2023-08-25
Bug fix when saving collection tree for first time in split repos

ryanmitchell:
Fixes https://github.com/statamic/eloquent-driver/issues/198

When saving a collection for the first time in split config mode (file for collection config, eloquent for data), it will return a [stache tree as the collection structure is designed to do this](https://github.com/statamic/cms/blob/e66df876e11cd8e757eed4fd0691380870e6d99b/src/Structures/CollectionStructure.php#L59C1-L60C1).

The work around is to not assume we get passed an eloquent collection tree.

### DIFF
--- a/src/Structures/CollectionTreeRepository.php
+++ b/src/Structures/CollectionTreeRepository.php
@@ -20,13 +20,15 @@ class CollectionTreeRepository extends StacheRepository
         });
     }
 
-    public function save($entry)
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Closing in favour of https://github.com/statamic/eloquent-driver/pull/203

Issue #201, opened 2023-09-06
GlobalVariables not imported (eloquent:import-globals)


maartenvanhunsel:
When importing globals our global data is not imported because `$set-localizations()` is an empty array. (see https://github.com/statamic/eloquent-driver/blob/master/src/Commands/ImportGlobals.php#L68C4-L69C16)

Is there any config file we haven't setup correctly? Or where does Statamic gets all these localizations?

PR #202, opened 2023-09-06
Fix issue with importing global variables

Ken-vdE:
Fix the issue when importing Globals from files to the database, where 'global variables' are incorrectly read from the database (which is obviously empty because we haven't imported yet) instead of the files.

Fixes https://github.com/statamic/eloquent-driver/issues/201


ryanmitchell:
Lovely thank you.

Issue #206, opened 2023-09-21
Redundant Caching in Asset.php's ->exists() Method Causes Missing Metadata During Asset Import


SylvesterDamgaard:
I'm encountering caching issues with the ->exists() method in the Asset.php file. I have some code that imports media files from our old CMS, but it seems to be missing metadata.

```php
$container = \Statamic\Facades\AssetContainer::find('images');
$asset = Assets::find($container->handle().'::'.$path);
if (!$asset){
  $asset = $container->makeAsset($path);
  $asset->upload($file);
  $asset->save();
}
```

I noticed that the generateMeta() method, which is called during the save process, doesn't properly handle the metadata section. The issue seems to be that the file list used for comparison is cached and doesn't get updated or cleared, even when media files are saved or deleted.

The method includes the following code:
```
$files = Blink::once($this->container()->handle().'::files', function () {
    return $this->container()->files();
});
```

The data returned from `$this->container()->files()` / `$this->container()->contents()` is already cached and gets updated during the save/delete process for assets.

I believe the caching in the Asset.php file is redundant and should be removed.

As a temporary workaround, I can call `Blink::forget($container->handle().'::files');` before `$asset->save()`, and metadata is successfully added as expected.

ryanmitchell:
Closed by https://github.com/statamic/eloquent-driver/pull/218

PR #207, opened 2023-09-21
Remove redundant cache

SylvesterDamgaard:
Fixes https://github.com/statamic/eloquent-driver/issues/206

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -49,15 +49,11 @@ class Asset extends FileAsset
 
     public function exists()
     {
-        $files = Blink::once($this->container()->handle().'::files', function () {
-            return $this->container()->files();
-        });
-
         if (! $path = $this->path()) {
             return false;
         }
 
-        return $files->contains($path);
+        return $this->container()->files()->contains($path);
     }
 
     private function metaValue($key)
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Thanks for this. I do understand that for your use case (importing) the extra cache is redundant, but for normal day to day use the extra cache does result in performance improvements which is why it was added.

I'm going to sit on this for a bit and see if I can come up with a way that solves both problems.

SylvesterDamgaard:
@ryanmitchell, I completely understand your concerns. However, I genuinely feel that using an immutable data strategy for dynamic data is inherently flawed, even though it's only applied on a per-process or per-request basis. I initially discovered this issue due to performance bottlenecks, as missing metadata on images can lead to significantly slower page loads or even timeouts.

As for workarounds, using `php please asset:meta` is also quite slow, and this becomes increasingly problematic as the asset count grows into the thousands. Clearing the cache, as per my example, is another option but it's not ideal.

Before submitting this PR, I conducted performance tests using a real dataset of over 40,000 images. My tests showed no noticeable changes in performance since the data is already being delivered from the cache. Do you have any test results that indicate otherwise?

I hope this clarifies my position and I'm open to further discussions to find the best solution.

ryanmitchell:
Yes, we have a similar number of assets on our live sites using this add-on. As I said, we noticed verifiable speed improvements which is why this was added. However there have been changes in statamic/cms that may have duplicated some of this work. I need sufficient time to investigate so it doesnt cause a regression - you can always bind your own Asset class to get around this.

SylvesterDamgaard:
@ryanmitchell Got it. Thank you for the clarification.

Issue #191, opened 2023-07-24
Please don't load all asset infos , when call exists method of the Statamic\ Eloquent\ Assets\Asset


videni:
Generate thumbnail for one image , it will  get  all image infos from aws s3, or alioss.

``
http://localhost:8000/cp/thumbnails/YXNzZXRzOjphX2ltb2xhX2NlcmFtaWNhX3N0aWxlX3N0aWxlX2dsXzYyNGNmLndlYnA=/small
``

which is extreamly slow, because it try to get all image infoes from remote aws s3, or alioss,  as you can see 
![image](https://github.com/statamic/cms/assets/4058341/fae8cffe-37bf-4fd0-8754-5b9134bde09b)

I have 50 thousand images, generating one thumbnai would halt my server for a long time.

ryanmitchell:
This is the same approach taken by the file driver - https://github.com/statamic/cms/blob/f8fcb062488507c2dc96e96b8a2377ce8c5c7246/src/Assets/Asset.php#L200-L207

Due to the possibility of having containers file based and assets not (or the other way around) we can't do any clever database optimisations to check for existence.

I agree it's an issue, I just dont think it can be solved by this driver. I think you should open this issue over on statamic/cms and link to this conversation.

videni:
@ryanmitchell Yes, you are right, Maybe the AssetContainer should add a exists method.

ryanmitchell:
Just looping back to this. I suppose you could also bind your own Asset model that extends the eloquent one and overwrite the exists method yourself? Maybe make it query the database instead of the file system? You can define your asset in the config file.

```php

namespace App\Eloquent;

class Asset extends \Statamic\Eloquent\Assets\Asset {
   public function exists() { ... }
}

```

videni:
@ryanmitchell , it is far more complex than what I expetected, lots of places will trigger the all methd of the AssetContainerContents class, the exists method is just one of them. I am afraid this would  not be resolved soon, most likely , it would take years.

ryanmitchell:
Yes it will fetch and cache data, but it also indexes it, similar to a database (the Stache is basically a flat file database). But this driver bypasses the Stache and uses eloquent.

If you are hitting issues with performance using the eloquent driver we should definitely investigate ways to resolve those as long as we preserve compatibility with the behaviour from the Stache driver.

SylvesterDamgaard:
[...]

PR #218, opened 2023-11-09
Make AssetQueryBuilder for performance gains

ryanmitchell:
This PR brings an AssetQueryBuilder to eloquent driver, reducing the number of file systems calls made and removing the need for a stache cache of the path listing.

This is particularly useful for sites with a large number of assets in cloud storage, as asides from initially populating a folder cache and accessing assets individually, all queries, filtering and meta will be only accessed from the database.

Replaces https://github.com/statamic/eloquent-driver/pull/207
Closes https://github.com/statamic/eloquent-driver/issues/191

### DIFF
--- a/database/migrations/updates/update_assets_table.php.stub
+++ b/database/migrations/updates/update_assets_table.php.stub
@@ -35,8 +35,7 @@ return new class extends Migration {
                 $model->basename = $path->afterLast('/');
                 $model->extension = Str::of($model->basename)->afterLast('.');
                 $model->filename = Str::of($model->basename)->beforeLast('.');
-                $model->meta = Arr::except($model->data, ['data']);
-                $model->data = $model->data['data'] ?? [];
+                $model->meta = $model->data;
                 $model->save();
             });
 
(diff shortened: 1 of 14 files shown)


database/migrations/updates/update_assets_table.php.stub
```diff
@@ -0,0 +1,67 @@
...
+            $table->string('path')->after('extension')->index();
+            $table->jsonb('meta')->after('path')->nullable();
+
+            $table->index(['container', 'folder', 'basename']);
+        });
+
+        AssetModel::all()
+            ->each(function ($model) {
+                $model->container = Str::before($model->handle, '::');
+                $model->folder = Str::of($model->handle)->after('::')->replace('/.meta/', '')->beforeLast('/');
```
SylvesterDamgaard, line 27: The value saved is not correct.

SylvesterDamgaard (changes requested):

database/migrations/updates/update_assets_table.php.stub
ryanmitchell: Thanks, I've updated the logic to handle root folder assets

SylvesterDamgaard:
I can confirm the migration is now working as expected.

Issue #223, opened 2023-12-11
Array to string conversion during eloquent sync


GioChocolateBro:
I just installed the 3.0 update. I have a sizeable filesystem with lots of folders in digitalocean spaces.

After installation the `eloquent:sync-assets` command throws an exception when it reaches the end of processing the root folder and trying to move on to nested folders.

```
local.ERROR: Array to string conversion {"exception":"[object] (ErrorException(code: 0): Array to string conversion at /var/www/html/vendor/laravel/framework/src/Illuminate/Support/Stringable.php:33)
[stacktrace]
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(254): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Array to string...', '/var/www/html/v...', 33)
#1 /var/www/html/vendor/laravel/framework/src/Illuminate/Support/Stringable.php(33): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->Illuminate\\Foundation\\Bootstrap\\{closure}(2, 'Array to string...', '/var/www/html/v...', 33)
#2 /var/www/html/vendor/laravel/framework/src/Illuminate/Support/Str.php(68): Illuminate\\Support\\Stringable->__construct(Array)
#3 /var/www/html/vendor/statamic/cms/src/Support/Str.php(314): Illuminate\\Support\\Str::of(Array)
#4 /var/www/html/vendor/statamic/eloquent-driver/src/Assets/AssetContainerContents.php(75): Statamic\\Support\\Str::__callStatic('of', Array)
#5 [internal function]: Statamic\\Eloquent\\Assets\\AssetContainerContents->Statamic\\Eloquent\\Assets\\{closure}(Array, 0)
#6 /var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Arr.php(869): array_filter(Array, Object(Closure), 1)
#7 /var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Collection.php(387): Illuminate\\Support\\Arr::where(Array, Object(Closure))
#8 /var/www/html/vendor/statamic/eloquent-driver/src/Assets/AssetContainerContents.php(76): Illuminate\\Support\\Collection->filter(Object(Closure))
[...]

Issue #222, opened 2023-12-07
Undefined array key "type"


crackAT:
Hi all, 

i switched to the new Asset Eloquent Container but there is always this error when accessing the asset browse file page: 

```
[2023-12-07 12:13:47] local.ERROR: Undefined array key "type" {"userId":1,"exception":"[object] (ErrorException(code: 0): Undefined array key \"type\" at /vendor/statamic/cms/src/Stache/Stores/ContainerAssetsStore.php:84)
[stacktrace]
#0 /vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(254): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Undefined array...', '/Users/mkrammer...', 84)
#1 /vendor/statamic/cms/src/Stache/Stores/ContainerAssetsStore.php(84): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->Illuminate\\Foundation\\Bootstrap\\{closure}(2, 'Undefined array...', '/Users/mkrammer...', 84)
#2 /vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php(851): Statamic\\Stache\\Stores\\ContainerAssetsStore->Statamic\\Stache\\Stores\\{closure}(Array, 'images')
#3 [internal function]: Illuminate\\Support\\Collection->Illuminate\\Support\\Traits\\{closure}(Array, 'images')
#4 /vendor/laravel/framework/src/Illuminate/Collections/Arr.php(871): array_filter(Array, Object(Closure), 1)
#5 /vendor/laravel/framework/src/Illuminate/Collections/Collection.php(387): Illuminate\\Support\\Arr::where(Array, Object(Closure))
#6 /vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php(853): Illuminate\\Support\\Collection->filter(Object(Closure))
#7 /vendor/statamic/cms/src/Stache/Stores/ContainerAssetsStore.php(89): Illuminate\\Support\\Collection->reject(Object(Closure))
#8 /vendor/statamic/cms/src/Stache/Stores/ContainerAssetsStore.php(68): Statamic\\Stache\\Stores\\ContainerAssetsStore->getFiles()
#9 /vendor/statamic/cms/src/Stache/Stores/ContainerAssetsStore.php(50): Statamic\\Stache\\Stores\\ContainerAssetsStore->paths()
[...]

PR #224, opened 2023-12-12
Fix issues with directory listing not returning all folders

ryanmitchell:
This PR fixes an issue with the sync command and directory listing using the asset query builder. It also adds metaExists caching to the asset class to prevent multiple duplicate queries.

Closes https://github.com/statamic/eloquent-driver/issues/222
Closes https://github.com/statamic/eloquent-driver/issues/223

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -133,6 +133,8 @@ class Asset extends FileAsset
         }
 
         $model->save();
+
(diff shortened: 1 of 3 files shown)


SylvesterDamgaard:
I tested this PR and it works as expected now! Thank you @ryanmitchell

Issue #229, opened 2024-01-16
Revisions will not work after upgrading to Statamic v4.41.0


helloiamlukas:
Statamic v4.41.0 introduced the following bugfix: https://github.com/statamic/cms/pull/9190
Unfortunately this breaks the revision functionality when using the eloquent driver.

Revisions are now loaded via the `id` attribute: https://github.com/statamic/cms/pull/9190/commits/8c3275a97c22a9a6bfe58d076cc64435d8c2e7fd

This attribute is not saved when using the eloquent driver:
https://github.com/statamic/eloquent-driver/blob/6eccbb6d477df3de1759bd3a70a0fdc85e2ee9a8/src/Revisions/Revision.php#L53

@ryanmitchell What's the reason for this? Can we add the `id` to the attributes without any further sideeffects?

ryanmitchell:
I've opened https://github.com/statamic/eloquent-driver/pull/230 to start storing id again. Im not sure why we weren't in the first place.

Unfortunately due to how the change has been made in core any old revisions will now be broken.

helloiamlukas:
Great, thank you! :)

> Unfortunately due to how the change has been made in core any old revisions will now be broken.

What do you think about making a migration/command, that will extract the id from the key and add it to the attributes?

helloiamlukas:
Done! https://github.com/statamic/eloquent-driver/pull/231

PR #231, opened 2024-01-17
Update old broken revisions after updating to Statamic v4.41.0

helloiamlukas:
Fixes https://github.com/statamic/eloquent-driver/issues/229 for existing revisions

### DIFF
--- a/database/migrations/updates/add_id_to_attributes_in_revisions_table.php.stub
+++ b/database/migrations/updates/add_id_to_attributes_in_revisions_table.php.stub
@@ -13,7 +13,7 @@ return new class extends Migration {
                 foreach ($revisions as $revision) {
                     $id = str($revision->key)
                         ->afterLast('/')
-                        ->toInteger();
+                        ->toString();
 
                     $attributes = $revision->attributes;
                     $attributes['id'] = $id;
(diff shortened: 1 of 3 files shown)


database/migrations/updates/add_id_to_attributes_in_revisions_table.php.stub
```diff
@@ -0,0 +1,42 @@
...
+    public function up()
+    {
+        // Extract the id from the key and add it to the attributes of the revision
+        RevisionModel::query()
+            ->whereJsonDoesntContainKey('attributes->id')
+            ->chunk(200, function ($revisions) {
+                foreach ($revisions as $revision) {
+                    $id = str($revision->key)
+                        ->afterLast('/')
+                        ->toInteger();
```
ryanmitchell, line 16: This would break when not using integer ids for entries - can you change it?

helloiamlukas: Ah yes, thanks for the notice! Will change it.

helloiamlukas:
Too late now, but it seems using `chunk` seems to cause some weird behaviour in Postgres. You would need to use `chunkById`, as recommended in the Laravel [docs](https://laravel.com/docs/10.x/queries#chunking-results).

Should I open another pull request?

helloiamlukas:
Alright, here we go: https://github.com/statamic/eloquent-driver/pull/232

PR #234, opened 2024-01-31
Check if a asset extension could be found

jonassiewertsen:
It may be, that I had corrupt assets in my import.

To not abort the import script, I added a check, if the extension could be found.
In case it returns null, the import aborts.

```
Integrity constraint violation: 1048 Column 'extension' cannot be null
```

To improve the check, some kind of notice, that a certain image could not be imported may be nice.
Anyways, I thought I'll drop this little change to see, if this is something that should be implemented.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -127,6 +127,11 @@ class Asset extends FileAsset
             $meta = $source->meta();
         }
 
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Looks good. I guess the only other approach would be to just set extension to an empty string? I guess we need to decide if a lack of extension is valid or not... ?

PR #248, opened 2024-02-21
Apply syncOriginal for dirty state tracking

jasonvarga:
This goes along with statamic/cms#5502 to allow `isDirty` etc to work correctly when using the Eloquent driver.

- [x] Entries
- [x] Terms
- [x] Assets
- [x] Trees
- [x] Composer requirement
- [x] Tests

### DIFF
--- a/tests/Data/Entries/EntryQueryBuilderTest.php
+++ b/tests/Data/Entries/EntryQueryBuilderTest.php
@@ -39,7 +39,7 @@ class EntryQueryBuilderTest extends TestCase
     public function entry_is_found_within_all_created_entries_and_select_query_columns_are_set_using_entry_facade_with_find_method_with_columns_param()
     {
         $searchedEntry = $this->createDummyCollectionAndEntries();
-        $columns = ['title'];
+        $columns = ['foo', 'collection'];
         $retrievedEntry = Entry::query()->find($searchedEntry->id(), $columns);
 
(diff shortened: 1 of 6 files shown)


jasonvarga:
The changes in the tests were because collections are technically required, although in the simple test cases they never got to the real world usage where collection would be used, so they passed.

Now that syncOriginal is being called, it looks for the collection and caused an error. These changes add collections where necessary.

tests/Data/Entries/EntryQueryBuilderTest.php
```diff
@@ -39,7 +39,7 @@ public function entry_is_found_within_all_created_entries_using_entry_facade_wit
     public function entry_is_found_within_all_created_entries_and_select_query_columns_are_set_using_entry_facade_with_find_method_with_columns_param()
     {
         $searchedEntry = $this->createDummyCollectionAndEntries();
-        $columns = ['title'];
+        $columns = ['foo', 'collection'];
```
jasonvarga, line 42: It's irrelevant for this particular test, since it's just about making sure `$columns` are reflected in `->selectedQueryColumns()`, but without selecting a collection it would throw an error. I also changed title to foo to show that the field itself is irrelevant.

jasonvarga:
Tests will break until Statamic 4.51.0 is released.

PR #249, opened 2024-03-04
Allow publishing migrations separately

duncanmcclean:
I'm building a new `php please install:eloquent-driver` command into Statamic Core, which will make the process of moving "things" to the Eloquent Driver much easier.

The command will install the package & let you configure which repositories you want to move over. 

As part of these changes, I didn't want to force users to publish *all* of the Eloquent Driver's migrations, even if they're only planning on switching one of two of the repositories.

This PR makes that possible by splitting the migrations into separate publish tags, grouped by repository.


ryanmitchell:
Made a couple of minor changes, let me know what you think.

duncanmcclean:
Changes look good to me. Thanks!

PR #250, opened 2024-03-06
Publish tree migrations under their own tag

duncanmcclean:
This pull request moves the migration for the `trees` table to its own publish tag, to allow for it to be published independently of the `navigations` table.

This PR also renames the migration to simply `create_trees_table` since the table is called `trees` and not `navigation_trees`. Feel free to rename this back if you think it'll cause issues.

Builds on the changes introduced in #249.


duncanmcclean:
> I wonder if we should take the chance to switch to a more conventional migrations approach and have them with fixed dates?

Yeah, I think that might be a better approach. It's the way Laravel handles it with their first-party packages. I'll close this PR and make that change in a separate PR.

PR #253, opened 2024-03-07
Use fixed timestamps in migration filenames

duncanmcclean:
This pull request changes the way migrations are published to prevent the same migration from being published multiple times.

Up until now, the Eloquent Driver has been generating the destination filename based on the current date & time. However, this causes an issue if a developer tries to re-run the `vendor:publish` command for the same tags since it won't detect the existing migration because the filename would be different.

I considered using Laravel's new `publishesMigrations` method (introduced in laravel/framework#49246), however it has the same pitfall in that it doesn't check for the existence of the same migration published at another date/time.


ryanmitchell:
Seems good. Do you want it merged or is the plan to hold off for v5?

duncanmcclean:
> Seems good. Do you want it merged or is the plan to hold off for v5?

Yeah, let's get it merged, it'll allow for testing the Core PR. Just wanted to check it looked good for you before merging since I've haven't touched this package much 😄

marvinosswald:
This causes issues with existing installations as my system now tries to migrate all of those again as the name changed from `2024_02_07...` to `2024_03_07...` am i overseeing something i should do differently ?

potsky:
Me too @marvinosswald.

The migrations seem to be loaded from the package now. So I have removed all targeted migrations from our repository and have renamed the migrations in the `migrations` table with this PostgreSQL query:

```sql
UPDATE migrations
SET migration='2024_03_07_100000' || SUBSTRING (migration, '^2024_01_09_[0-9]{6}(_create.*$)')
WHERE migration LIKE '2024_01_09_%_create_%_table';
```

Change the `2024_01_09` date with your installation date of course!

@duncanmcclean Can you confirm we need to do this?

duncanmcclean:
When are the migrations being re-published? When you run `composer update`?

marvinosswald:
https://github.com/statamic/eloquent-driver/pull/253/files#diff-abf1fa77a19f720052b75bc802df0260157b9064d8c3c12a80794f01e9acc065L216

Doesn't need to be republished for this problem.
this is the problematic line, this makes laravel "discover" the migrations without them being published afaik.

duncanmcclean:
> Doesn't need to be republished for this problem.
this is the problematic line, this makes laravel "discover" the migrations without them being published afaik.

That line is only related to loading the migrations within the test suite, it won't affect "real" applications. 

I'm asking since I'm unable to replicate the issue myself, on a site using the Eloquent Driver. Updating to the latest version doesn't result in migrations being published so trying to figure out what could be causing this 🤔

duncanmcclean:
Ah, I've figured it out...

In addition to letting you publish the migrations yourself into your `database/migrations` directory, the Eloquent Driver is also directly loading the migration files from it's `database/migrations` directory, which leads to errors when running `php artisan migrate`. I'll tag a release to fix this shortly.

Issue #258, opened 2024-03-13
Assets gets duplicated when there is at least one capital letter in the name


shawinigan:
We try tu upload assets from the CP but they gets inserted in double in the database when there's a capital letter in the name of the file.  We just renamed the file and here's the behaviour : 

billethackqc.pdf  => working fine
billethackqc1.pdf => working fine
billethackqc_1.pdf => working fine
billethackqc_111111111111111111111111111111111111111111111111111111111.pdf => working fine
billethackqc_1Allo.pdf => get inserted in double

Again, it was the same pdf file, we only renamed it and got this behaviour. It seems any file with a capital letter is having the same issue.

ryanmitchell:
Whats the value of `statamic.config.assets.lowercase` ?

shawinigan:
'lowercase' => true,

ryanmitchell:
Thanks, I'll spent some time later in the week seeing if I can replicate this.

ryanmitchell:
I don't seem to be able to replicate this using a local filesystem. What filesystem are you using?

modrictin:
Hey there,

If I try to save an asset which contains a whitespace to `assets_meta` table, I get the following error:
```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'one-public-im-az-1710414015.png' for key 'assets_meta.assets_meta_container_folder_basename_unique' (Connection: mysql, SQL: update `assets_meta` set `basename` = im-az-1710414015.png, `filename` = im-az-1710414015, `path` = public/im-az-1710414015.png, `meta` = {"data":[],"size":861322,"last_modified":1710414015,"width":897,"height":673,"mime_type":"image\/png","duration":null}, `assets_meta`.`updated_at` = 2024-03-14 11:00:15 where `id` = 20)
```

I also get 2 entries saved to my database after the error:
![image](https://github.com/statamic/eloquent-driver/assets/31577031/0b517c79-26e6-4405-97e7-1248b8b78612)

Maybe it is connected to this.

Thanks in advance,
Tin

ryanmitchell:
@modrictin can you verify you are on the latest version (3.3.2) as I'm not seeing this behaviour there. How are you creating assets?

shawinigan:
> I don't seem to be able to replicate this using a local filesystem. What filesystem are you using?

Here's the config in the filesystem along with the part related to it in the .env file

 'documents' => [
[...]

Issue #251, opened 2024-03-07
Assets with normalized characters duplicated on Upload


jstoks:
Assets with filenames that include characters that will be replaced by sanitization will cause a duplicate record in the `assets_meta` table. When viewing the index of the assets container there will be a row for the original filename with 0 size and timestamp.

I tested this using the local file system and S3. They both have this issue.

Additionally, this phantom file causes the asset index to slow down to a crawl (25 seconds) when using S3. Deleting the erroneous assets_meta record fixes the slowdown. 

### Example

In this example I'm uploading a file with spaces in the filename.

![Screenshot 2024-03-06 at 6 05 06 PM](https://github.com/statamic/eloquent-driver/assets/2073157/6c45478f-aada-479d-ae2a-667063a6ea80)

```
-[ RECORD 1 ]-------
id         | 27
container  | image_assets
folder     | /
basename   | 2020 07 26.jpeg
filename   | 2020 07 26
extension  | jpeg      
path       | 2020 07 26.jpeg
meta       | {"data": []}
created_at | 2024-03-07 00:16:02
updated_at | 2024-03-07 00:16:02
-[ RECORD 2 ]-------
id         | 28
container  | image_assets
folder     | /
basename   | 2020-07-26.jpeg
filename   | 2020-07-26
extension  | jpeg      
path       | 2020-07-26.jpeg
meta       | {"data": [], "size": 149857, "width": 680, "height": 510, "duration": null, "mime_type": "image/jpeg", "last_modified": 1709770562}
created_at | 2024-03-07 00:16:02
updated_at | 2024-03-07 00:16:02
```


### Support Details

```
Environment                                                                                                                                     
Application Name: Laravel                                                                                                                       
Laravel Version: 10.46.0                                                                                                                        
PHP Version: 8.1.27                                                                                                                             
Composer Version: 2.7.1                                                                                                                         
Environment: local                                                                                                                              
Debug Mode: ENABLED                                                                                                                             
URL: localhost                                                                                                                                  
Maintenance Mode: OFF                                                                                                                           
                                                                                                                                                
Cache                                                                                                                                           
Config: NOT CACHED                                                                                                                              
Events: NOT CACHED                                                                                                                              
Routes: NOT CACHED
Views: NOT CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: pgsql
Logs: cli_stack / background, bugsnag
Mail: smtp
Queue: sync
Scout: database_index
Session: file

Statamic
[...]

PR #259, opened 2024-03-15
Don't write meta when we should only be generating it

ryanmitchell:
This PR fixes https://github.com/statamic/eloquent-driver/issues/258#issuecomment-1997579873 and https://github.com/statamic/eloquent-driver/issues/251 in a different manner, by going to the root cause of the extra meta rows.

This was being caused by overly aggressive caching (cache vs blink) and the generateMeta method writing meta, which again was a cache optimisation that we have had to roll back.

@jstoks @modrictin @shawinigan I've run this with all the filenames you provided and it works ok for me - can you pull the changes down and verify they work in your environments?

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -3,17 +3,14 @@
 namespace Statamic\Eloquent\Assets;
 
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Support\Facades\Cache;
 use Statamic\Assets\Asset as FileAsset;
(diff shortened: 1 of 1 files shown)


jstoks:
I appreciate the follow up, however we decided to remove Statamic from our application. I won't be able to test this.

modrictin:
Hey @ryanmitchell ,

sorry for the late reply.

This works for me and it has resolved all the issues.

Thanks,
Tin

PR #267, opened 2024-04-10
Statamic 5: Simplify `TestCase`

duncanmcclean:
This pull request simplifies the addon's `TestCase`, thanks to statamic/cms#9573.

I've also removed the custom `partialMock` method which no longer seems to be needed and the `assertContainsHtml` assertion which we don't seem to be using anywhere.


ryanmitchell:
hmm since this has merged a lot of the tests arent performing any assertions:

CleanShot 2024-04-11 at 06 32 30@2x

duncanmcclean:
I _think_ that may be related to the PHPUnit version. I'll take a look today.

PR #277, opened 2024-04-26
Tokens support

jacksleight:
Adds token support so live preview tokens are stored in the database.

There are a couple of things that are out of the ordinary:

- There's no `TokenRepository` contract, so we're just binding to the file repository class instead.
- The file token `make` method doesn't use the service container, so we have to override it.
- Calling `app(Token::class)::fromModel($model)` doesn't work due to the constructor arguments, so this just calls `Token::fromModel($model)` directly.

These probably require PRs to core if we want them to work like everything else.

Let me know if anything needs changing.

### DIFF
--- a/src/Tokens/Token.php
+++ b/src/Tokens/Token.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Statamic\Eloquent\Tokens;
+
+use Illuminate\Database\Eloquent\Model;
+use Statamic\Contracts\Tokens\Token as Contract;
+use Statamic\Tokens\Token as FileToken;
+
+class Token extends FileToken
+{
+    protected $model;
+
+    public static function fromModel(Model $model)
+    {
+        return (new static($model->token, $model->handler, $model->data))
+            ->expireAt($model->expire_at)
+            ->model($model);
+    }
+
+    public function toModel()
+    {
+        return self::makeModelFromContract($this);
+    }
+
+    public static function makeModelFromContract(Contract $source)
+    {
+        $class = app('statamic.eloquent.tokens.model');
+
+        return $class::firstOrNew(['token' => $source->token()])->fill([
+            'handler'   => $source->handler(),
+            'data'      => $source->data(),
+            'expire_at' => $source->expiry(),
+        ]);
+    }
+
+    public function model($model = null)
+    {
+        if (func_num_args() === 0) {
+            return $this->model;
+        }
+
+        $this->model = $model;
+
+        return $this;
+    }
+
+    public function save()
+    {
+        $model = $this->toModel();
+        $model->save();
+
+        $this->model($model->fresh());
+    }
+
+    public function delete()
+    {
+        $this->model()->delete();
+    }
+}
--- a/src/Tokens/TokenRepository.php
+++ b/src/Tokens/TokenRepository.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Statamic\Eloquent\Tokens;
+
+use Statamic\Contracts\Tokens\Token as TokenContract;
+use Statamic\Tokens\TokenRepository as StatamicTokenRepository;
+
+class TokenRepository extends StatamicTokenRepository
+{
+    public function make(?string $token, string $handler, array $data = []): Token
+    {
+        return new Token($token, $handler, $data);
+    }
+
+    public function find($token)
+    {
+        $model = app('statamic.eloquent.tokens.model')::whereToken($token)->first();
+
+        if (! $model) {
+            return;
+        }
+
+        return Token::fromModel($model);
+    }
+
+    public function save($entry)
+    {
+        $model = $entry->toModel();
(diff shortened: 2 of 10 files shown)


SylvesterDamgaard:
@ryanmitchell PR for supporting the preview tokens as we talked about. :)

src/Updates/CreateTokensTable.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Statamic\Eloquent\Updates;
```
duncanmcclean, line 3: This should probably be `UpdateScripts` instead, to stay inline with Core.

```suggestion
namespace Statamic\Eloquent\UpdateScripts;
```

ryanmitchell: We've used updates in eloquent driver (since before my time), so its consistent internally at least.

duncanmcclean: Oh okay! Didn't realise that. 👍

ryanmitchell:
Awesome, thanks for this.

I've added a PR to core for the cli install support: https://github.com/statamic/eloquent-driver/pull/277

We're switching to default repositories `file` rather than `eloquent` from the next release, so I've gone ahead and changed that in advance of that. You can just run the migration manually anyway, and opt-in (which is what I would expect on our installs).

jasonvarga:
> There's no TokenRepository contract, so we're just binding to the file repository class instead.

This was dumb on my part. Since this PR is targeting Statamic 5, we can get a contract into core that you can bind to.

src/Tokens/TokenRepository.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Statamic\Eloquent\Tokens;
+
+use Statamic\Contracts\Tokens\Token as TokenContract;
+use Statamic\Tokens\TokenRepository as FileTokenRepository;
+
+class TokenRepository extends FileTokenRepository
```
jasonvarga, line 8: You should be able to continue extending `Statamic\Tokens\TokenRepository`, but it has no "file" logic in it so maybe just change `as FileTokenRepository` to `as Repository`.

src/Tokens/TokenRepository.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    public function make(?string $token, string $handler, array $data = []): Token
+    {
+        return new Token($token, $handler, $data);
+    }
```
jasonvarga, line 13: You shouldn't need to override `make` anymore.

src/Tokens/Token.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace Statamic\Eloquent\Tokens;
+
+use Illuminate\Database\Eloquent\Model;
+use Statamic\Contracts\Tokens\Token as Contract;
+use Statamic\Tokens\Token as FileToken;
+
+class Token extends FileToken
```
jasonvarga, line 9: Continue extending `Token` but probably just alias it to `AbstractToken`.

src/Tokens/Token.php
```diff
@@ -0,0 +1,59 @@
...
+        $model = $this->toModel();
+        $model->save();
+
+        $this->model($model->fresh());
+    }
+
+    public function delete()
+    {
+        $this->model()->delete();
+    }
```
jasonvarga, line 58: These aren't needed. The abstract methods will pass the token to the repository's save/delete methods where you've already added this logic.

jasonvarga (changes requested):
Once statamic/cms#9964 is merged, you should be able to do the mentioned changed.

Also, you'll probably need to adjust typehints.

ryanmitchell:
All looks good to me, but I'll let Jason merge it as he was working with you on this.

src/Tokens/TokenRepository.php
```diff
@@ -0,0 +1,42 @@
...
+        $model = app('statamic.eloquent.tokens.model')::whereToken($token)->first();
+
+        if (! $model) {
+            return null;
+        }
+
+        return Token::fromModel($model);
+    }
+
+    public function save(TokenContract $token): bool
```
jacksleight, line 21: I did run into an issue here (and in `delete`). This hint should really be `Statamic\Eloquent\Tokens\Token`, since it uses `toModel`, but if I set it to that it complained it wasn't compatible with the base class. Not sure if I was doing it wrong or PHP was just being weird.

jasonvarga: `Statamic\Eloquent\Tokens\Token` does implement `TokenContract` so that's fine. You can call `toModel` on it. I think maybe just your IDE would be complaining that the `toModel` method didn't exist since it's not in the contract. Should be fine though.

You can probably add a docblock to get that IDE support.

```php
/**
 * @param \Statamic\Eloquent\Tokens\Token $token
 */
public function save(TokenContract $token): bool
```

PR #278, opened 2024-04-26
Feature/add protect to entry model

JelleGroenendal:
We wanted to use the protect feature that the MD files have but this doesn't work by default in the Eloquent solution.
With these changes it does work.


ryanmitchell:
Thanks! I'd probably be inclined to just use the data column for it, much like how blueprint works. Saves a migration?

If you want to continue with the migration thats fine, but it would need an update script and the base migration would need updated (currently your update migration doesn't run).

JelleGroenendal:
I agree on that. I moved this way considering this is how the MD files work. I removed the migration and updated the model to use data protect

ryanmitchell:
Can you supply a test for this where it doesn't work without this change? For me adding `protect` to the data column already works without this change.

JelleGroenendal:
Your correct i cant believe i missed that it already worked with the data field

PR #279, opened 2024-05-01
Fix querying statuses

duncanmcclean:
This pull request fixes an issue I ran into this morning when using the REST API alongside the Eloquent Driver.

Essentially, when the REST API queries to get entries, it filters for entries where `status` `is` `published`. However, when using the Eloquent Driver, no entries were being returned. 

I've fixed this issue by copying over the recent changes around entry statuses to the Eloquent Driver (related: #9317). 

The implementation is *mostly* the same as it is in Core. However, the main difference is the fact the Eloquent Driver's query builder doesn't maintain a `$collections` property to track the queried collection, instead we're just using `$this->wheres()`

## To Do

* [x] Get the tests passing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     },
     "require": {
         "php": "^8.1",
-        "statamic/cms": "v5.0.0-alpha.6"
+        "statamic/cms": "dev-where-status-trait"
     },
     "require-dev": {
         "doctrine/dbal": "^3.3",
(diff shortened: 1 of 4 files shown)


jasonvarga:
When building the status stuff in core, I was thinking that Eloquent driver extended that. I didn't realize we would need to copy over `whereStatus` and all the related methods.

I'm making a PR into core to extract it into a trait that'll make this PR much leaner.

PR #285, opened 2024-05-10
Drop foreign key on form_id

ryanmitchell:
As reported on discord there is an error in the form migration, we needed to drop the foreign key first.

This PR ensures that is done.


CasEbb:
Where would this foreign key have been created? This PR is breaking our upgrade v3 -> v4 right now.

ryanmitchell:
@CasEbb can you check if these updates fix the issue on your side? https://github.com/statamic/eloquent-driver/pull/324

Issue #286, opened 2024-05-15
Unable to store form submission


andreas-eisenmann:
Scenario: create a form and set store submissions to true.

Now we have this call stack:

```
\Statamic\Http\Controllers\FormController::submit
|_ \Statamic\Eloquent\Forms\Submission::save
    |_ \Statamic\Eloquent\Forms\Submission::toModel
```
In `toModel()` we have:

```php
$model = $class::findOrNew($this->id());
```
The result of `$this->id()` is a timestamp - no wonder, just look at the code:

https://github.com/statamic/cms/blob/7cbc505da6c7f2f9568599631b607aa427d5cc62/src/Forms/Submission.php#L52

```php
return $this->id = $id ?: str_replace(',', '.', microtime(true));
```
Surprise, now we have a database exception:

```
SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type bigint: "1715767251.4272"
CONTEXT:  unnamed portal parameter $1 = '...' (Connection: pgsql, SQL: select * from "form_submissions" where "form_submissions"."id" = 1715767251.4272 limit 1)
```

How is this supposed to work?

ryanmitchell:
It seems to be due to Postgres being less forgiving than other engines - see the comments here: https://github.com/statamic/eloquent-driver/pull/122#issuecomment-2111965758

PR #287, opened 2024-05-15
Change ID type on form_submissions to be a float

ryanmitchell:
This PR updates the id type on form_submissions to be a float, rather than an int, so that out of the box it works on Postgres.

@andreas-eisenmann would you mind checking that this migration works in your situation?

Closes https://github.com/statamic/eloquent-driver/issues/286

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "statamic/cms": "^5.0.1"
     },
     "require-dev": {
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.5",
         "laravel/pint": "^1.0",
         "orchestra/testbench": "^8.0 || ^9.0.2",
         "phpunit/phpunit": "^9.4 || ^10.0 || ^11.0"
--- a/database/migrations/2024_05_15_100000_modify_form_submissions_id.php
+++ b/database/migrations/2024_05_15_100000_modify_form_submissions_id.php
@@ -16,7 +16,8 @@ return new class extends Migration
     public function down()
     {
         Schema::table($this->prefix('form_submissions'), function (Blueprint $table) {
-            $table->id('id')->change();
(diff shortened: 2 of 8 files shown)


andreas-eisenmann:
@ryanmitchell thanks! First try after executing the migration fails with:

```
SQLSTATE[23502]: Not null violation: 7 ERROR: null value in column "id" of relation "form_submissions" violates not-null constraint DETAIL: Failing row contains (null, {"date": "2024-05-15T11:42:30.516004Z", "email": "andreas.eisenm..., 2024-05-15 13:42:30.516004, 2024-05-15 13:42:30.519891, 1)
```

I will start debuging now to find the root cause for this error.

andreas-eisenmann:
Now with your changes in https://github.com/statamic/eloquent-driver/commit/9a37ad8cef71eac4b6c851dc621fe3cafd7474a8 it works. 👍️

lazerg:
Hi @ryanmitchell,

After updating, this code stopped to work.

I migrated new migration but anyway it is not working?

image

Error:

```
local.ERROR: SQLSTATE[HY000]: General error: 1364 Field 'id' doesn't have a default value (Connection: mysql, SQL: insert into `statamic_form_submissions` (`form`, `data`, `updated_at`, `created_at`) values (success-stories, {"email":"lazerg2@gmail.com","companyName":"myCompany","firstName":"Lazizbek","date":"2024-06-27T01:49:07.017736Z"}, 2024-06-27 11:49:07.018737, 2024-06-27 11:49:07.018737))
```

ryanmitchell:
@lazerg if you are make()-ing the model directly you'll need to specify an id, it no longer auto increments.

Instead you should make on the Facade:

```php

\Statamic\Facades\Form::find('success-stores')->makeSubmission()
	->data([...])->save();
```

Issue #204, opened 2023-09-15
Split blueprints for forms


royduin:
Currently we did set the blueprints driver to "file" and disable the permissions to change them on all environments (through https://github.com/justbetter/statamic-environment-permissions) so the end user can not make blueprint changes. This is great as the customer should not make any blueprint changes to collections, taxonomies, globals, etc as changing it could case the frontend to break. And if the customer wants to add something; that should probably be visible at the frontend so a developer is needed to handle it. But.. a form blueprint is more like content. 

We made it dynamic with: https://github.com/rapidez/statamic/blob/master/resources/views/page_builder/form.blade.php so the customer should be able to create their own forms.

Can we split this? So form blueprints are stored in the database and the rest in a file? Probably something with permissions also need to change within Statamic itself?

ryanmitchell:
Closing as it a duplicate of https://github.com/statamic/eloquent-driver/issues/164

royduin:
But that's about form submissions, not the form blueprint.

ryanmitchell:
You can have all blueprints in the database or all in file. There is no distinction between the type of blueprint.

royduin:
Yeah, that's the point. We'd like to split that so we can version control the blueprints of collections, globals, etc and have the form blueprints in the database as that's more like content.

ryanmitchell:
Yep I understand - thats not currently possible. I'll open this back up, but being honest I'm not sure if/how it could be done.

royduin:
Just to keep track of it and maybe see if other users want this as well. If we find some time we'll dig into this further.

ryanmitchell:
With recent Statamic versions there is a separate permission for form blueprint editing to regular blueprint editing. Does this solve the problem for you?

royduin:
That's nice but we also need that separation here: https://github.com/statamic/eloquent-driver/blob/master/config/eloquent-driver.php#L19-L23 so form blueprints can be stored in the database and all others can stay in file.

ryanmitchell:
I don't think that's something this driver should support. 

That said, you could do it by binding your own BlueprintRepository and amending the in() method to check for file by default, and use eloquent if the namespace is form.

 https://github.com/statamic/eloquent-driver/blob/efaa724dc49ca7c8f90af8aa46ddecbb6b560c4a/src/Fields/BlueprintRepository.php#L63

ryanmitchell:
[...]

PR #300, opened 2024-06-12
Support mixed repository blueprints

ryanmitchell:
This PR adds a config to choose which blueprint namespaces are eloquent driven and which are not. For example, you may want form blueprints in the database but others to be file based.

Closes https://github.com/statamic/eloquent-driver/issues/204


jasonvarga:
Nah

PR #301, opened 2024-06-14
make use of $ids in updateEntryOrder method

mkwia:
This PR optimises the collection query in the `updateEntryOrder` method by making use of `ids` which are redundant in the current method.

On some of our sites this caused a large number of jobs to be dispatched to the queue when entries were reordered in collection trees. This fix has resolved our issue when applied via composer patch.

The fix is akin to logic already in use in the `updateEntryUris` method

### DIFF
--- a/src/Collections/CollectionRepository.php
+++ b/src/Collections/CollectionRepository.php
@@ -82,8 +82,11 @@ class CollectionRepository extends StacheRepository
 
     public function updateEntryOrder(CollectionContract $collection, $ids = null)
     {
-        $collection->queryEntries()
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Wonderful, thank you. I've changed it slightly to use `->when` as it reads better (to me).

clementmas:
I'm getting errors from this PR: `Undefined variable $query`.

I think this should be `->when($ids, fn ($query) => $query->whereIn('id', $ids))`

ryanmitchell:
@clementmas thanks - fixed in v4.3.1

PR #302, opened 2024-06-17
Allow blueprints and field sets to be split repository

ryanmitchell:
This PR allows blueprints and field sets to be run in split repository mode, ie the blueprints could be eloquent driven and field sets file driven, or vice versa.

It also fixes some of the code introduced in https://github.com/statamic/eloquent-driver/pull/300 to handle methods that were missed initially.

### DIFF
--- a/config/eloquent-driver.php
+++ b/config/eloquent-driver.php
@@ -42,7 +42,7 @@ return [
     ],
 
     'fieldsets' => [
-        'driver' => 'eloquent',
+        'driver' => 'file',
         'model' => \Statamic\Eloquent\Fields\FieldsetModel::class,
     ],
 
--- a/config/eloquent-driver.php
+++ b/config/eloquent-driver.php
@@ -18,8 +18,7 @@ return [
 
     'blueprints' => [
         'driver' => 'file',
-        'blueprint_model' => \Statamic\Eloquent\Fields\BlueprintModel::class,
-        'fieldset_model' => \Statamic\Eloquent\Fields\FieldsetModel::class,
+        'model' => \Statamic\Eloquent\Fields\BlueprintModel::class,
         'namespaces' => 'all',
     ],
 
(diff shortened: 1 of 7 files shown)


config/eloquent-driver.php
```diff
@@ -18,8 +18,7 @@
 
     'blueprints' => [
         'driver' => 'file',
-        'blueprint_model' => \Statamic\Eloquent\Fields\BlueprintModel::class,
-        'fieldset_model' => \Statamic\Eloquent\Fields\FieldsetModel::class,
+        'model' => \Statamic\Eloquent\Fields\BlueprintModel::class,
```
duncanmcclean, line 21: Since these repositories are being split up, they should probably become separate options in the `php please install:eloquent-driver` command.

config/eloquent-driver.php
```diff
@@ -42,6 +41,11 @@
         'entry' => \Statamic\Eloquent\Entries\Entry::class,
     ],
 
+    'fieldsets' => [
+        'driver' => 'eloquent',
```
duncanmcclean, line 45: This should be `file` by default, like the other repositories.

```suggestion
        'driver' => 'file',
```

Otherwise, as soon as you install the Eloquent Driver, it'll attempt to get fieldsets from the database even if you haven't migrated the table, causing an error.

duncanmcclean (changes requested):

config/eloquent-driver.php
ryanmitchell: Good spot, forgot to change it when testing

duncanmcclean (approved):

src/ServiceProvider.php
```diff
@@ -3,6 +3,7 @@
 namespace Statamic\Eloquent;
 
 use Illuminate\Foundation\Console\AboutCommand;
+use JetBrains\PhpStorm\Deprecated;
```
jasonvarga, line 6: This seems like a mistake. (You started using PhpStorm? 😄 )

ryanmitchell: definitely a mistake.

i told you i'd give it a go... i don't hate it as much as i thought i would :)

ryanmitchell: i think it added it when i added `@deprecated` in a comment

jasonvarga:
This is not push-back at all, but I'm just curious why someone would want blueprints in the db and fieldsets in files (and vice-versa).

jasonvarga:
Ah yeah I remember that conversation. Thanks.

PR #305, opened 2024-06-19
Use double instead of float for MySQL compatability

ryanmitchell:
Change the form id migration to use double as float isn't consistent in MySQL


clementmas:
FYI, Laravel's `double()` function doesn't accept a precision parameter:

```php
/**
 * Create a new double column on the table.
 *
 * @param  string  $column
 * @return \Illuminate\Database\Schema\ColumnDefinition
 */
public function double($column)
{
    return $this->addColumn('double', $column);
}
```

So it will use the default MySQL DOUBLE precision which is around 15 digits, which should still work I think.

ryanmitchell:
My mistake, it was meant to be decimal not double

PR #310, opened 2024-06-26
Handle vendor namespaces and handle for blueprints and fieldsets

ryanmitchell:
When importing add-on namespaced field sets and blueprints from the stache the namespace (for blueprints) is prefixed with vendor, and for field sets the handle is prefixed too. 

This PR updates the logic to handle those appropriately.

### DIFF
--- a/src/Commands/ImportBlueprints.php
+++ b/src/Commands/ImportBlueprints.php
@@ -139,6 +139,11 @@ class ImportBlueprints extends Command
             $handle = Str::before($basename, '.yaml');
             $handle = str_replace('/', '.', $handle);
 
+            // handle any add-on fieldsets
+            if (Str::startsWith($handle, 'vendor.')) {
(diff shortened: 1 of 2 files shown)


lazerg:
Hi @ryanmitchell 

After this update, fieldsets that are exported by vendor:publish from third-party packages are broken

Screenshot 2024-08-20 at 13 38 07

ryanmitchell:
@lazerg what is the handle of fieldset in the database?

lazerg:
@ryanmitchell 

after update it is similar to second one format, but `statamic/cms` looking for first one format

image

ryanmitchell:
@lazerg does this PR fix the issue for you? https://github.com/statamic/eloquent-driver/pull/339

ryanmitchell:
Thanks for confirming, I'll bundle it into the next release.

Issue #312, opened 2024-06-26
Unable to order select options in a blueprint when using eloquent driver


Baumi-Dev:
### Bug description

When using eloquent driver, the form input options are stored as JSON data, which doesn't allow for specific ordering. Options should be stored in an array to allow for specific ordering

### How to reproduce

Ordering options in flat files works fine, but when using eloquent driver, options are stored in a JSON object, causing the options to be stored in ascending order. Reordering options in the control panel doesn't work, as the options are always stored in ascending order

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 11.10.0
PHP Version: 8.2.10
Composer Version: 2.7.2
Environment: local
Debug Mode: ENABLED
URL: empower-eats.public.test
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.11.0 PRO

Statamic Addons
statamic/eloquent-driver: 4.5.0

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: file
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

ryanmitchell:
Can you provide a screencast or screenshot of what you are seeing and also the underlying data please?

Baumi-Dev:
Hi Ryan, 

Thanks for getting back to me so quickly.

Please find screen recording and screenshot attached.

Cheers,

Alex
[Uploading Archive.zip…]()

Baumi-Dev:
Here's a screencast as well...

https://github.com/statamic/eloquent-driver/assets/138436068/eeae327b-3fde-4d80-a626-9df3e050f90b

ryanmitchell:
Thanks - just for clarity it's blueprint related rather than form related. 

[...]

PR #321, opened 2024-07-16
Preserve select field order in blueprints and fieldsets

ryanmitchell:
This PR updates the logic when saving blueprints and fieldsets to ensure that the order of select field options is preserved. Currently we rely on the underlying database engine to preserve the order which is unreliable.

- [x] Blueprints
- [x] Fieldsets

Closes https://github.com/statamic/eloquent-driver/issues/312

### DIFF
--- a/src/Fields/BlueprintRepository.php
+++ b/src/Fields/BlueprintRepository.php
@@ -9,6 +9,8 @@ use Statamic\Support\Arr;
 
 class BlueprintRepository extends StacheRepository
 {
+    use Traits\StoresAndRetrievesFieldOrder;
+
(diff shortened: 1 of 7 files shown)


Baumi-Dev:
Awesome @ryanmitchell! Cheers mate, I'll test it out

ryanmitchell:
I think this is going to be addressed by a change in how these fields are stored in core: https://github.com/statamic/cms/pull/10467

So I'm going to hold fire and see how that plays out.

ryanmitchell:
Going to close this as Jason has solved it over in core.

PR #328, opened 2024-07-24
Drop indexes before dropping the column

georgecoca:
SQLite requires deleting the indexes before deleting the column. Relates to this: https://github.com/laravel/framework/pull/51373


ryanmitchell:
Lovely, thank you.

PR #330, opened 2024-07-31
Only retrieve forms once instead of for every item

dnwjn:
Hello everyone!

We stumbled upon a bug on `/admin/forms`. We would get endless loading screens and eventually timeouts when visiting this page. We have quiet an amount of form submissions (~150.000), and the query builder in the current setup retrieves all submissions, with the form for every one of them.

We're using Statamic 5 with this package, and even though we were able to fix this in the `FormsController` of the `statamic/cms` package (see https://github.com/statamic/cms/pull/10534), we figured this could still cause problems for other scenarios outside of the `FormsController`. So we wanted to propose it.

The proposed changes retrieve all forms only once before being connected to each submission, instead of retrieving the form every time. This drastically decreases the amount of queries being executed, and it fixes the timeout issues we were dealing with.

If you have any questions or comments about this, let me know!

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 use Statamic\Contracts\Forms\Form as Contract;
 use Statamic\Events\FormDeleted;
 use Statamic\Events\FormSaved;
+use Statamic\Facades\Blink;
 use Statamic\Forms\Form as FileEntry;
 
 class Form extends FileEntry
@@ -60,6 +61,9 @@ class Form extends FileEntry
 
         $this->model($model->fresh());
 
+        Blink::forget("eloquent-forms-{$this->handle()}");
+        Blink::forget('eloquent-forms');
+
         FormSaved::dispatch($this);
     }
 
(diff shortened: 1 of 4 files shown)


ryanmitchell:
I would prefer we blink the query result in the Form Repository, similar to how we do it elsewhere (eg collections - https://github.com/statamic/eloquent-driver/blob/47d3b3af84ff5da1074509c3ca4f7f46a365c400/src/Collections/CollectionRepository.php#L12-L26) 

Do you want to try refactoring it to take that approach and see how it works from a performance perspective on your install?

dnwjn:
@ryanmitchell Appreciate the quick response! I've implemented Blink, and I'd say it gives about the same results (~6.5 seconds). I see how Blink is more in line with the rest of the package, so let me change that right away!

ryanmitchell:
I've pushed up some additions to clear the blink cache on save/delete as per the collection repository, and I've also added test coverage as we're trying to get the test suite up to a satisfactory level.

Thanks again for the PR, appreciate you taking the time to submit it.

dnwjn:
@ryanmitchell Thanks for taking care of this so quickly! If I ever create another PR I'll make sure to take testing into account as well!

PR #345, opened 2024-08-27
Attempt to make term <> collection query more performant

ryanmitchell:
This PR updates the collection checking clause in the TermQueryBuilder to be more performant.

Previously we were looping over all entries to get their terms, and when the number of entries increases this became an issue. Now, we loop over the terms and check if an entry has that term, and if so we query for it, otherwise we don't.

The hope is that most users will have less terms than entries. 

Ideally we'd have a pivot table of some sort but that would require significant changes in core.

### DIFF
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -223,6 +223,7 @@ class TermQueryBuilder extends EloquentQueryBuilder
                         }
 
                         return TermModel::where('taxonomy', $taxonomy)
+                            ->select('slug')
                             ->get()
                             ->map(function ($term) use ($taxonomy) {
                                 return Entry::query()
@@ -232,7 +233,6 @@ class TermQueryBuilder extends EloquentQueryBuilder
                             })
                             ->filter()
                             ->values();
-
                     });
 
                     if ($terms->isNotEmpty()) {
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -215,46 +215,34 @@ class TermQueryBuilder extends EloquentQueryBuilder
                     ? Taxonomy::handles()->all()
                     : $this->taxonomies;
 
-                // get entries in each collection that have a value for the taxonomies we are querying
-                // or the ones associated with the collection
-                // what we ultimately want is a subquery for terms in the form:
-                // where('taxonomy', $taxonomy)->whereIn('slug', $slugArray)
-                $collectionTaxonomyHash = md5(collect($this->collections)->merge(collect($taxonomies))->sort()->join('-'));
+                collect($taxonomies)->each(function ($taxonomy) use ($query) {
+                    $collectionTaxonomyHash = md5(collect($this->collections)->merge([$taxonomy])->sort()->join('-'));
 
-                Blink::once("eloquent-taxonomy-hash-{$collectionTaxonomyHash}", function () use ($taxonomies) {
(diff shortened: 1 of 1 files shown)


michaeljhopkins:
This update works great. Big performance improvements from creating >20k models down to ~20ish

jonatanalvarsson:
@ryanmitchell I realize of course that it's the retrieving of all related entries (when augmenting the `$entry->taggar` property) that takes time here. I can't use `$entry->get('taggar')` since that will only get me the raw term slugs. Is there a simple way to get the augmented term data _without_ loading all related entries for every term?

Btw, I don't use Antlers on this project at all - I send augmented data to Vue using Inertia.

ryanmitchell:
It’s a bit of a no win, which is worse scenario. We either query the terms to get the entries (like now) or query the entries to get the terms (as before). The theory is that there will be more people the new approach will benefit. 

Really the only way to get the best of both worlds is a relationship table but that would require changes in core. 

If you aren’t using antlers you could get() the data and then use the term facade to do the lookup (with a wherein) rather than relying on augmentation?

jonatanalvarsson:
@ryanmitchell Thanks for your quick reply!

> If you aren’t using antlers you could get() the data and then use the term facade to do the lookup (with a wherein) rather than relying on augmentation?

Yeah, sure - for this specific project it's not super hard for me to work around this issue since I have full control over augmenting my data before sending it to the frontend. But A) that's not the case for everyone and B) in the control panel it now takes 2.5 seconds to load 21 entries if the taxonomy term field is included - which is a remarkable increase from 189 ms in version 4.13.

However. After some additional testing I found a "fix" for my specific case. When I [detached](https://statamic.dev/taxonomies#collections) the taxonomy "taggar" from the collection "articles" (but still kept the taxonomy terms field "taggar" in my blueprint) response times and query counts decreased enormously (on both site and cp).

### v4:13 (before this PR) and 1 taxonomy attached
entry with 0 terms: 55ms and 6 queries (23 models)
entry with 1 term: 75ms and 12 queries (27 models)

### v4.14 (after this PR) and 1 taxonomy attached
entry with 0 terms: **2.4s** and **3434** queries (3424 models) 😢
entry with 1 term: **525ms** and **600** queries (595 models)

### v4.14 (after this PR) and 0 taxonomies attached
entry with 0 terms: 45ms and 4 queries (2 models) 😃 
entry with 1 term: 52ms and 5 queries (3 models)

It still feels like the code changes from version 4.13 to 4.14 could use some additional testing. For collections with taxonomies attached the new code does seem to produce an insane amount of queries if you have a lot of terms. When loading an entry with one term it makes one query per taxonomy term in the attached taxonomy (589 in my case). And when loading an entry with zero terms it makes one query per taxonomy term for all taxonomies (3422 in my case). At least that last thing still feels like a bug.

ryanmitchell:
> It still feels like the code changes from version 4.13 to 4.14 could use some additional testing. For collections with taxonomies attached the new code does seem to produce an insane amount of queries if you have a lot of terms. When loading an entry with one term it makes one query per taxonomy term in the attached taxonomy (589 in my case). And when loading an entry with zero terms it makes one query per taxonomy term for all taxonomies (3422 in my case). At least that last thing still feels like a bug.

Yes unfortunately this is to be expected as its what happens in core too - unfortunately neither are bugs but are expected behaviour due to the need to match the behaviour in core. Ideally we'd just do a whereNotNull on the taxonomy value, but thats not quite the same. Maybe we just accept that difference @jasonvarga ?

PR #352, opened 2024-09-12
The root folder is / and not empty otherwise it's skipped

danielsmink:
The root folder of the container is listed as `/` in the database. Since the default was an empty string the default folder was never synced. For us this resulted in a big list of assets in assets meta that were no longer present on disk causing false 404's on pages.

This PR sets folder to `/` by default so the root folder (of the container) is also synced.


ryanmitchell:
Lovely, thank you.

Issue #359, opened 2024-10-03
Import of assets freezes


faltjo:
### Bug description

When moving asset containers and assets to eloquent, the import of the assets is freezing at 0%.
I can not reproduce the issue with a minimal example unfortunately. Currently we have around 26k assets and two containers in production.

I was able to fix the issue, but I have no clue why it works like this. Maybe it it the interplay between `Statamic\Eloquent\Assets\AssetContainer` and `Statamic\Assets\Asset` (since `$this->queryAssets()` returns objects of the latter type)?

Change this line

https://github.com/statamic/eloquent-driver/blob/995c09da6d42b8d11e68591ac09f7d475cc7131c/src/Assets/AssetContainer.php#L140

to

```
    ->transform(fn($value) => $value->path());
```

and the import works.

### How to reproduce

1. Create some AssetContainers and Assets without Eloquent:
`config('eloquent-driver.asset_containers.driver')='file'` and `config('eloquent-driver.assets.driver')='file'`
2. Set the driver of AssetContainers and Assets to Eloquent: `config('eloquent-driver.asset_containers.driver')='eloquent'` and `config('eloquent-driver.assets.driver')='eloquent'`
3. Import the existing AssetContainers and Assets to your DB: `php please eloquent:import-assets`
4. Importing the AssetContainers works flawlessly, however import of Assets freezes at 0% progress.

### Environment

```yaml
Environment
Laravel Version: 11.22.0
PHP Version: 8.3.12
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Locale: de

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: pusher
Cache: redis
Database: pgsql
Logs: stack / daily
Mail: ses
Queue: sync
Session: file

Livewire
Livewire: v3.4.12

Statamic
Addons: 4
Sites: 1
Stache Watcher: Disabled
Static Caching: Disabled
Version: 5.24.0 PRO

Statamic Addons
rias/statamic-address-field: 1.3.0
rias/statamic-redirect: 3.7.1
statamic-rad-pack/meilisearch: 3.3.0
statamic/eloquent-driver: 4.14.1

Statamic Eloquent Driver
Asset Containers: eloquent
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: file
Entries: eloquent
Forms: file
Global Sets: file
[...]

PR #360, opened 2024-10-03
Prevent freezing of ImportAssets for asset meta data when importing a large number of files

faltjo:
fixes https://github.com/statamic/eloquent-driver/issues/359


faltjo:
Feel free to remove the test if you think it is superfluous.

Issue #361, opened 2024-10-04
Exporting Entries creates new IDs


lokmanm:
### Bug description

Hey folks, we are trying to revert a site to Flat Files, and when Exporting Entries new `ids` are created. 

This breaks every relationship within the site.

Is there any way to disable this or do we need to do a PR?

Also Collection Trees are not exported at all, tried a couple of times. 

### How to reproduce

Try to move back to Flat Files from DB and check IDs.

### Environment

```yaml
Application Name: Statamic
Laravel Version: 11.22.0
PHP Version: 8.3.12
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Drivers
Broadcasting: null
Cache: file
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: file

Statamic
Addons: 5
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.24.0 PRO

Statamic Addons
alt-design/alt-redirect: 1.3.2
doefom/statamic-export: 0.3.2
rias/statamic-data-import: 1.5.0
statamic/eloquent-driver: 4.14.1
statamic/seo-pro: 6.1.1

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: file
```

lokmanm:
Okay if anyone needs this and you are in a hurry this is what we did:

We copied the ExportEntries command to our project.

We added the following function to include the `id` for entries:

```php
$entry = Entry::make()
        ->id($model->id)
        ...rest of the func
```

Now go to `vendor/statamic/eloquent-driver/src/Entries/EntryModel.php` and add the following to the model:

```php
protected $keyType = 'string';
```

This will make sure the IDs are set to `string` as for some reason they are set to `int`. 

This change will keep the same `IDs` as the SQL data.

Happy to submit a PR if the Core Team agrees on this.

ryanmitchell:
I think the exporter should use the same entry model you specify in the config - and if that’s a got a UUID trait then it should bring over the ids.

lokmanm:
[...]

PR #362, opened 2024-10-04
Fix entries export to use correct model, blueprint column, and set ID when exporting UUIDs

ryanmitchell:
This PR updates the export entries logic to use the model specified in the config, rather than forcing a numeric ID based model. This allows us to check if its a UUID model and if so also to export out IDs.

@lokmanm this is what I was meaning.

Closes https://github.com/statamic/eloquent-driver/issues/361


lokmanm:
@ryanmitchell yes, that does it!

Issue #373, opened 2024-11-04
Optimize taxonomy queries.


TheBnl:
### Description

Writing this here to check if this is wanted behaviour or not. The issue I was facing is that on a project with 100+ tags I saw the total amount of page queries reflecting that amount. Diving deeper into where these queries where coming from I traced it back to the TermQueryBuilder. In the method `applyCollectionAndTaxonomyWheres` that is used when retrieving page taxonomies, a map is run over all taxonomies to check when they are used on the page. This creates, in my case, 134 count queries being fired on every page request. The result is only cached in the blink request cache.  

The current query:
```php
$query = TermModel::where('taxonomy', $taxonomy)
    ->select('slug')
    ->get()
    ->map(function ($term) use ($taxonomy) {
        return Entry::query()
            ->whereIn('collection', $this->collections)
            ->whereJsonContains('data->'.$taxonomy, [$term->slug])
            ->count() > 0 ? $term->slug : null;
    })
    ->filter()
    ->values();
```

The reason is, probably, that someone can configure the Entries to be stored with the file driver. 

My suggestion would be to check what entry driver is configured and based on that opt for a join to come to the result reducing the 133 queries to 1.

Suggestion:
```php
if (config('statamic.eloquent-driver.entries.driver', 'file') == 'eloquent') {
    $query = TermModel::where('taxonomy', $taxonomy)
        ->join('entries', function (JoinClause $join) use ($taxonomy) {
            $join->on('entries.collection', '=', 'entries.collection')
                ->where('entries.collection', $this->collections)
                ->whereRaw("json_contains(`entries`.`data`, concat('[\"',`taxonomy_terms`.`slug`,'\"]'), '$.\"$taxonomy\"')");
        })
        ->select('taxonomy_terms.slug')
        ->get();
} else {
    $query = TermModel::where('taxonomy', $taxonomy)
        ->select('slug')
        ->get()
        ->map(function ($term) use ($taxonomy) {
            return Entry::query()
                ->whereIn('collection', $this->collections)
                ->whereJsonContains('data->'.$taxonomy, [$term->slug])
                ->count() > 0 ? $term->slug : null;
        })
        ->filter()
        ->values();
}
```
I must admit that this join is a bit funky, especially the on statement `'entries.collection', '=', 'entries.collection'` but it gets the job done. Definitely open to suggestions to optimize or replace it with something better.
But in the end the N+1 query issue could definitely be solved.

ryanmitchell:
[...]

PR #376, opened 2024-11-09
Optimize taxonomy <> collection queries

TheBnl:
Working on a draft to fix issue #373 

The current implementation does not yet pass the tests due to not being fully db driver independent. 
The test fails because `JSON_QUOTE` seems to behave different in sqllite than mysql.
The `JSON_QUOTE` should probably not be hard coded here as well because we can't assume the db driver that is in use.

Closes https://github.com/statamic/eloquent-driver/issues/373

### DIFF
--- a/src/Taxonomies/TermQueryBuilder.php
+++ b/src/Taxonomies/TermQueryBuilder.php
@@ -226,23 +226,23 @@ class TermQueryBuilder extends EloquentQueryBuilder
                             return [];
                         }
 
+                        // workaround to handle potential n+1 queries on the database site
+                        // if/when Statamic core supports relationships in a meaningful way this should be removed
                         if (config('statamic.eloquent-driver.entries.driver', 'file') == 'eloquent') {
                             $entriesTable = (new EntryModel)->getTable();
                             $termsTable = (new TermModel)->getTable();
+
                             $query = TermModel::where('taxonomy', $taxonomy)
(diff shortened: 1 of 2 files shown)


ryanmitchell:
@TheBnl let's move the conversation over here. I've modified the approach and have tests working on SQLite and it works fine on MySQL from my own testing.

Do you have access to sqlserver or postgres to check this on?

TheBnl:
> @TheBnl let's move the conversation over here. I've modified the approach and have tests working on SQLite and it works fine on MySQL from my own testing.
> 
> Do you have access to sqlserver or postgres to check this on? It also needs some test coverage if you could work on that.

Awesome! Will test on postgress, haven't came by to installing sqlserver yet. Will check and look into adding postgress/sqlserver to the github test workflow. :+1:

TheBnl:
Tested with postgres and there we do need wrap it so the value will be handled as json: `to_jsonb({$wrappedColumn}::text)`

Working on the test coverage in a different branch, also to test sqlserver,  as it is quite a bit of work of settings things up right.

ryanmitchell:
Sorry I didn't mean we need to set up different testing environments for each database engine - I just meant we needed a test in the suite to cover this new behaviour.

TheBnl:
Ah haha well that makes things quite simpler. I've added a test where the entries driver is set to 'file' so both the join and map gets tested.

ryanmitchell:
Thanks for your work on this - really appreciate it.

TheBnl:
Awesome thanks for merging!

PR #379, opened 2024-11-18
Update entry URIs when collection route is changed

### LINKED ISSUE #378: Update entry URI's when collection route is changed
### Description

I recently had to change the route of my collection, and I found that due to using eloquent-driver, all the entries of that collection became inaccessible afterwards, as the uri no longer matched the one in entries table

Perhaps a simple command will do the trick?
Claude.ai wrote this one

```
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use Statamic\Facades\Entry;

class UpdateEntryUris extends Command
{
    protected $signature = 'statamic:update-uris';
    protected $description = 'Update entry URIs in database';

    public function handle()
    {
        $entries = Entry::query()->where('collection', 'erlebnisse')->get();
        
        foreach ($entries as $entry) {
            $entry->save(); // This will trigger URI regeneration
            $this->info("Updated URI for entry: {$entry->id()}");
        }

        $this->info('All URIs updated successfully');
    }
}

```

duncanmcclean:
This pull request fixes an issue when changing a collection's route, where it wouldn't trigger the `uri` column on the `entries` table to be updated.

Closes #378.

### DIFF
--- a/src/Collections/CollectionRepository.php
+++ b/src/Collections/CollectionRepository.php
@@ -4,6 +4,7 @@ namespace Statamic\Eloquent\Collections;
 
 use Illuminate\Support\Collection as IlluminateCollection;
 use Statamic\Contracts\Entries\Collection as CollectionContract;
+use Statamic\Eloquent\Jobs\UpdateCollectionEntryUris;
 use Statamic\Facades\Blink;
 use Statamic\Stache\Repositories\CollectionRepository as StacheRepository;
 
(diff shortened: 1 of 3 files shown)


ryanmitchell:
This looks good but will only work if you have both collection and entries eloquent driven. 

If you have collections file driven, then this wouldn't update the eloquent entries with the new route (which I'd guess is the more common config).

Maybe we are ok with that, but thought I'd mention it for discussion.

duncanmcclean:
Oh, good point! 

Ideally, it'd work if your collections were stored as flat files too. I'll have a think about the best way to approach it 🤔

duncanmcclean:
Closing in favour of statamic/cms#11150.

PR #382, opened 2024-11-19
Apply and restore sort direction on asset containers

daun:
- Continuation of #381
- Read sort direction and field from config
- Restore into json


ryanmitchell:
Great, thanks.

PR #392, opened 2024-12-10
Ensure we store the latest Entry uri

ryanmitchell:
This PR updates entry saving to use the routableUri, not the uri, which ensures we always have the latest uri in the database, even when the entry is new or when the slug is changed.

Closes https://github.com/statamic/eloquent-driver/issues/391


kram-dyoll:
@ryanmitchell That worked as expected. The `slug` and `uri` fields were both updated at the same time and resulted in the new URL working first time.

PR #393, opened 2025-01-04
Adds Taxonomy export arguments, matching import

adampatterson:
I needed to export taxonomies but didn't need terms, I then saw that the import command comes with arguments that allows you to import just taxonomies or terms.

This PR adds the same arguments to the export command.

```shell
php please eloquent:export-taxonomies --only-taxonomies
php please eloquent:export-taxonomies --only-terms
```


ryanmitchell:
Thank you!

PR #395, opened 2025-02-11
Adjust length of form handle column in form submission table

Lenitr:
I ran into an issue while migrating form submissions from files to eloquent.

Some of the handles of my forms are longer than 30 characters, so i ran into an exception because there's currently a max length of 30 on the form column in the form submissions table.
In the forms table, this column has 255 characters. Not sure why the discrepancy is there (maybe because of the index on the column?)

I fixed my migration by adjusting the migration file for submissions like I did below.


ryanmitchell:
Thanks @Lenitr ! I mad a small tweak but this looks like a sensible change.

Lenitr:
> Thanks @Lenitr ! I mad a small tweak but this looks like a sensible change.

Ah I see, that defaults to 255. Thanks, I'll keep that in mind for the future 👍🏼

PR #396, opened 2025-02-12
return revisions in chronological order

faltjo:
Statamic revisions are expected to be returned in chronological order. 
When using file based revisions, this is always the case, since the revisions are written in there in correct order and read back line-by-line.

When using eloquent and for example postgres, the ordering of the returned rows is undefined and depends on postgres internals, therefore an explicit `orderBy` clause is introduced.

### DIFF
--- a/src/Revisions/RevisionRepository.php
+++ b/src/Revisions/RevisionRepository.php
@@ -16,6 +16,7 @@ class RevisionRepository extends StacheRepository
     public function whereKey($key)
     {
         return app('statamic.eloquent.revisions.model')::where('key', $key)
+            ->orderBy('created_at')
             ->get()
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Thanks for the PR. Can you tell me where in the CP this affects?

faltjo:
Sure, when using revisions open an entry and click 'Show History' (or something similar I do not know the exact translation) on the right panel. There the order of the revisions is depending on the order in which they are delivered by the database.

The revisions are loaded from here
https://github.com/statamic/cms/blob/5.x/src/Http/Controllers/CP/Collections/EntryRevisionsController.php#L16

ryanmitchell:
Great thanks, thats helpful for context.

PR #398, opened 2025-02-20
[5.x] Cast entry dates to UTC

duncanmcclean:
This pull request ensures that whenever a date is stored or retreived from the `date` column on the `entries` table, a UTC Carbon object will be returned.

Without our accessor, whenever the `date` attribute was accessed, the returned `Carbon` object would use the application's timezone, not UTC (which is what we want dates to be stored in instead).

Related: https://github.com/statamic/cms/pull/11409

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,6 @@ By default, the Eloquent Driver stores all data in a single `data` column. Howev
         {
             return [
                 // The casts from Statamic's base model...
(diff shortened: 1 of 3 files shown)


duncanmcclean:
Closing this as we've since changed our approach. Dates will continue to be saved in the app timezone, and only transformed to UTC when Statamic touches this.

PR #405, opened 2025-03-11
If template is present in data then make sure its assigned

ryanmitchell:


ryanmitchell:
@faustbrian not until the field sets issue is resolved

Issue #408, opened 2025-03-12
`lastModified()` on `LocalizedTerm` returning incorrect date


tanerkay:
### Bug description

Similar to #65, except this time for `Statamic\Taxonomies\LocalizedTerm`.

Bit more difficult to solve this time, as there is no `Statamic\Contracts\Taxonomies\LocalizedTerm` to create a new binding for.

The `lastModified()` method in `LocalizedTerm` is checking for `updated_at` as a timestamp, but in Eloquent, the `updated_at` is a Carbon instance. So what is happening is [the string version of the Carbon instance is being parsed into a timestamp](https://github.com/statamic/cms/blob/v5.50.0/src/Data/TracksLastModified.php#L13), which results in approximately 2100 seconds (Carbon adds the integer values of each part of the date string for some reason, e.g. `2025-03-12 10:50:35` results in `2025 + 3 + 12 + 10 + 50 + 35 = 2135` or 2135 seconds past the Unix epoch).

### How to reproduce

```php
use Statamic\Eloquent\Taxonomies\Taxonomy;
use Statamic\Facades\Term as TermFacade;
use Statamic\Taxonomies\LocalizedTerm;
    
Taxonomy::make('test')->title('test')->save();

tap(TermFacade::make('test-term')->taxonomy('test')->data([]))->save();

/** @var LocalizedTerm $term */
$term = TermFacade::query()->first();
$term->set('foo', 'bar');
$term->save();

$term->updated_at->toDateTimeString();
// example output: 1970-01-01 00:35:26
// note that actual time will depend on when the command is run
```

or as a test:

```php
use Statamic\Eloquent\Taxonomies\Taxonomy;
use Statamic\Facades\Term as TermFacade;
use Statamic\Taxonomies\LocalizedTerm;

#[Test]
public function it_saves_updated_at_value_correctly()
{
    $this->freezeSecond();

    Taxonomy::make('test')->title('test')->save();

    tap(TermFacade::make('test-term')->taxonomy('test')->data([]))->save();

    /** @var LocalizedTerm $term */
    $term = TermFacade::query()->first();
    $term->set('foo', 'bar');
    $term->save();

    $this->assertEquals(now(), $term->updated_at);
}
```

### Logs

```shell

```

### Environment

```yaml
Can be reproduced on a fresh clone of https://github.com/statamic/eloquent-driver
```

ryanmitchell:
Should be easily solved - https://github.com/statamic/eloquent-driver/issues/408

PR #409, opened 2025-03-12
Store updated_at in terms as timestamp

ryanmitchell:
Closes https://github.com/statamic/eloquent-driver/issues/408

### DIFF
--- a/src/Taxonomies/Term.php
+++ b/src/Taxonomies/Term.php
@@ -38,7 +38,7 @@ class Term extends FileEntry
         $term->data($data);
 
         if (config('statamic.system.track_last_update')) {
-            $term->set('updated_at', $model->updated_at ?? $model->created_at);
+            $term->set('updated_at', ($model->updated_at ?? $model->created_at)->timestamp);
         }
 
         return $term->syncOriginal();
--- a/src/Taxonomies/Term.php
+++ b/src/Taxonomies/Term.php
@@ -38,7 +38,9 @@ class Term extends FileEntry
         $term->data($data);
 
         if (config('statamic.system.track_last_update')) {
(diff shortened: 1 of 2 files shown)


src/Taxonomies/Term.php
```diff
@@ -38,7 +38,7 @@ public static function fromModel(Model $model)
         $term->data($data);
 
         if (config('statamic.system.track_last_update')) {
-            $term->set('updated_at', $model->updated_at ?? $model->created_at);
+            $term->set('updated_at', ($model->updated_at ?? $model->created_at)->timestamp);
```
tanerkay, line 41: Nice, only issue is that a future retrieval of a term, e.g. `TermFacade::query()->first()->updated_at` won't work as it's already an integer

Not sure if this is the most elegant way of solving it:
```suggestion
            $updatedAt = $model->updated_at ?? $model->created_at;

            $term->set('updated_at', $updatedAt instanceof Carbon ? $updatedAt->timestamp : $updatedAt);
```

ryanmitchell: No this is only called when a term is pulled from the database and made into a statamic term.

tanerkay: I could be doing something wrong then

I added `$this->assertEquals(now(), TermFacade::query()->first()->updated_at);` to the test and got an error. I didn't look at it for long though, I assumed something was cached. Sorry if I'm leading you on a wild goose chase!

ryanmitchell: Hmm ok not quite sure what's going on, something is being changed that shouldn't be in the LocalizedTerms. I'll apply your fix but I want to revisit it as somethings not right.

PR #411, opened 2025-03-18
[5.x] Separate globals config and content

### LINKED ISSUE #303: Globals not editable if entry is missing in database
### Bug description

Our goal is to keep globals definitions in the repository (driver: file) but store the global variables in a database.

When a new global set is added and deployed to a machine with a db that does not contain an entry in the `global_set_variables` table for it. Content manager have no possibility in the CP to edit, they only can access the edit page of the global set, the `/cp/globals/{slug}` page shows a 404.

The only way to make it editable seems to be running `php please eloquent:import-globals` but that overwrites all other variables in DB with data from the files which is undesired. There seem to be no way to populate only missing ones without overwriting data.

Am I missing something or do I use the driver in a wrong way?

### How to reproduce

Standard config, with only variables stored with eloquent driver
```php
    'global_sets' => [
        'driver' => 'file',
        'model' => \Statamic\Eloquent\Globals\GlobalSetModel::class,
    ],

    'global_set_variables' => [
        'driver' => 'eloquent',
        'model' => \Statamic\Eloquent\Globals\VariablesModel::class,
    ],
```

Create a global set

```
// delete item from table
php artisan migrate:fresh
```

### Logs

_No response_

### Environment

```yaml
Application Name: Statamic
Laravel Version: 11.10.0
PHP Version: 8.3.8
Composer Version: 2.7.7
Environment: local
Debug Mode: ENABLED
URL: localhost
Maintenance Mode: OFF

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: statamic
Database: sqlite
Logs: stack / single
Mail: log
Queue: sync
Session: file

Statamic
Addons: 1
Sites: 1
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.8.0 Solo

Statamic Addons
statamic/eloquent-driver: 4.3.0

Statamic Eloquent Driver
Asset Containers: file
Assets: file
Blueprints: file
Collection Trees: file
Collections: file
Entries: file
Forms: file
Global Sets: file
Global Variables: eloquent
Navigation Trees: file
Navigations: file
Revisions: file
Taxonomies: file
Terms: file
Tokens: file
```

### Additional details

_No response_

duncanmcclean:
This pull request makes the required changes to accommodate the changes in statamic/cms#11585, including:

* Added support for the new `sites` option on global sets
  * Statamic will now check against the sites in the `sites` array, rather than checking the existence of global variables in sites. 
* Drops the `origin` column on the `global_set_variables` table, as the origin now comes from the `sites` array.
* Added an update script ([`UpdateGlobalVariables`](https://github.com/statamic/cms/blob/a72607ad44e0bef02e470e7e688c6463df67c599/src/UpdateScripts/UpdateGlobalVariables.php)), mirroring the update script from my Core PR, but just touching the database rather than YAML files. 
* Updated some existing tests


duncanmcclean:
While the tests are failing on GitHub Actions, they all pass locally when using the `separate-globals` branch from `cms`.

PR #429, opened 2025-05-24
Add index to the date column

afonic:
This is a very simple PR to add an index to the `date` column of the `entries` table.

Sorting by the date column is a pretty common operation and it can be quite expensive without the index. In a low memory setup I would get a `QLSTATE[HY001]: Memory allocation error: 1038 Out of sort memory, consider increasing server sort buffer size` error even with just 60.000 entries, which is solved just by adding this index.

If there is another reason the date column is not indexed, sorry in advance!


ryanmitchell:
Thanks for this - I cant see any downside to adding it. 
I've added a migration for so its not applied to existing installs as well as new ones.

PR #433, opened 2025-05-29
Remove stache stores for repositories we control

ryanmitchell:
Requires https://github.com/statamic/cms/pull/11830


duncanmcclean:
Marking this as a draft until we figure out the CMS side.

Issue #445, opened 2025-06-08
(n+1) performance issue using assets


Jubeki:
### Bug description

Hey everyone,

I found a (n+1) performance issue in the handling of assets while using the eloquent driver.

The problem occurs, when you have multiple assets on the same field (in my example repo there are 3 assets).
This will result in 4 queries, where in my opinion the first query should be enough to fetch all assets (the other queries are fired to check if the assets meta exists, though they are already fetched in the first query).

```sql
select * from "assets_meta" where "container" = 'assets' and "path" in ('352-200x300.jpg', '990-200x300.jpg', '666-200x300.jpg')
select count(*) as aggregate from "assets_meta" where ("container" = 'assets' and "folder" = '/' and "basename" = '352-200x300.jpg')
select count(*) as aggregate from "assets_meta" where ("container" = 'assets' and "folder" = '/' and "basename" = '990-200x300.jpg')
select count(*) as aggregate from "assets_meta" where ("container" = 'assets' and "folder" = '/' and "basename" = '666-200x300.jpg')
```

How we encountered this issue:
We had a field with more than 20 assets, and saw degraded performance, so we dove deeper and saw n+a queries.

The Problem lies in `Statamic\Eloquent\Assets\Asset`, where the exists method is called by statamic:

```php
public function exists()
{
    return $this->existsOnDisk || $this->metaExists();
}

public function metaExists()
{
    return Blink::once('eloquent-asset-meta-exists-'.$this->id(), function () {
        return app('statamic.eloquent.assets.model')::query()
            ->where([
                'container' => $this->containerHandle(),
                'folder' => $this->folder(),
                'basename' => $this->basename(),
            ])->count() > 0;
    });
}
```

### How to reproduce

1. Setup a fresh statamic application which uses the eloquent-driver.
2. Create an Assets Field in the Page blueprint (allow multiple assets)
3. Upload multiple assets to the home page
4. Render the assets on the home page
5. Log all queries in the AppServiceProvider:

```php
DB::listen(function(QueryExecuted $query) {
    Log::debug($query->toRawSql());
});
```

6. Go to the home page to execute queries.
7. Take a look at `storage/logs/laravel.log` to see all queries.

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Statamic
Laravel Version: 12.17.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
URL: statamic-multiple-assets-query-issue.test
[...]

PR #446, opened 2025-06-08
Prevent n+1 queries on asset meta exists

ryanmitchell:
Closes https://github.com/statamic/eloquent-driver/issues/445

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -39,7 +39,7 @@ class Asset extends FileAsset
             ->syncOriginal();
 
         Blink::put('eloquent-asset-'.$asset->id(), $model);
-        Blink::put('asset-meta-'.$asset->id(), $model->meta);
+        Blink::put($asset->metaCacheKey(), $model->meta);
 
         return $asset;
     }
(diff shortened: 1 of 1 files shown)


src/Assets/Asset.php
```diff
@@ -32,11 +32,15 @@ public function syncOriginal()
...
     public static function fromModel(Model $model)
     {
-        return (new static)
+        $asset = (new static)
             ->container($model->container)
             ->path(Str::replace('//', '/', $model->folder.'/'.$model->basename))
             ->hydrateMeta($model->meta)
             ->syncOriginal();
+
+        Blink::put('asset-meta-'.$asset->id(), $model->meta);
```
Jubeki, line 41:
```suggestion
        Blink::put('eloquent-asset-'.$asset->id(), $model->meta);
```

It works with the above change (used a different key).

You could also use `Blink::put('eloquent-asset-meta-exists-'.$asset->id(), true)` and remove the additional `Blink::has('eloquent-asset-'.$this->id())` in metaExists

Jubeki:
```suggestion
        Blink::put($asset->metaCacheKey(), $model->meta);
```

`metaCacheKey` is a public method of Assets, so you could use this in a unified way:
https://github.com/statamic/cms/blob/eba307923eb5fe9ce7b0954bf6ceecc0357be8c1/src/Assets/Asset.php#L324-L327

Jubeki: Yeah, the reduction still works with the latest changes.

PR #455, opened 2025-06-12
Define descendant relationship on Entry model

BobWez98:
I have changed the way the directdescendants are queried as the default Statamic way causes an N + 1 issue for us. This reduces a lot of queries. 

The only problem still with this PR is the tests are failing. For some reason when creating entries in the tests it does not propagate them as they should. Does anyone have any idea? Or some pointers?

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -204,14 +204,14 @@ class Entry extends FileEntry
 
     public function directDescendants()
     {
-        return Blink::once('entry-descendants-' . $this->id(), function () {
+        return Blink::once('entry-descendants-'.$this->id(), function () {
             return new EntryCollection($this
-                    ->model()
-                    ->descendants()
-                    ->get()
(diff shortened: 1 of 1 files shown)


ryanmitchell:
Thanks for clarifying, though looking at your debug bar the change has made both the memory usage and response time higher. Less queries but a slower response?

I've updated your PR to resolve the descendants being cached on the model itself.

ryanmitchell:
As this is no longer finding entries through the repository/facade we lose the blink cache on the entries returned, which is why there are so many more models being returned. A lot of these will be duplicates, so this PR has replaced one inefficiency with another. I understand the desire to get the query count down, but getting the model count minimised is of equal importance.

As it stands, I don't think this solution is the correct general approach - it may work in your use case but it would have a detrimental effect in others. The good thing is that you can already apply this to your install by specifying your own Entry class and Model in the config.

PR #460, opened 2025-06-30
enable ordering by date range column

faltjo:
Currently ordering by a date column whose `mode` is set to `range` breaks with error
`SQLSTATE[22007]: Invalid datetime format: 7 ERROR:  invalid input syntax for type date: "{"end": "2025-06-30", "start": "2025-06-01"}" ...`

This PR enables such orderings: it orders by start date and when those values are equal additionally by end date, both in the requested direction.


faltjo:
Closed due to wrong target branch.

PR #470, opened 2025-07-08
Respect configured namespaces for blueprint import

macaws:
This PR adds the following enhancements:

- When importing blueprints, only the configured namespaces will be imported.
- A new option for `--all-blueprints` will skip this check for users that want to import them anyways (mimic existing functionality)
- Update readme to include namespace information

The only concern I have with this implementation is any nuance I may not be aware of - such as the default blueprint being excluded now since the namespace is `null` on it. Will let Statamic team provide input on this.

### DIFF
--- a/tests/Commands/ImportBlueprintsTest.php
+++ b/tests/Commands/ImportBlueprintsTest.php
@@ -118,7 +118,7 @@ class ImportBlueprintsTest extends TestCase
             ],
         ])->save();
 
-        BlueprintFacade::make('contact')->setNamespace('assets')->setContents([
+        BlueprintFacade::make('images')->setNamespace('assets')->setContents([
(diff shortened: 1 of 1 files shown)


ryanmitchell:
This looks good, and dont' worry about the default not coming over, it will fallback to the stache default.

Would you mind adding a test to https://github.com/statamic/eloquent-driver/blob/4.x/tests/Commands/ImportBlueprintsTest.php?

macaws:
> This looks good, and dont' worry about the default not coming over, it will fallback to the stache default.
> 
> Would you mind adding a test to https://github.com/statamic/eloquent-driver/blob/4.x/tests/Commands/ImportBlueprintsTest.php?

How about 2 tests 😉

PR #474, opened 2025-07-10
[4.x] Fix empty localized values not being set properly

Jade-GG:
Because the original code was only mapping existing values from the model data, when you unset a value in a specific localization that value didn't end up getting retrieved properly. This made it impossible to unset values in a localization that were set in the default site.

To fix this I turned around the code, mapping all of the `__localized_fields` to their model data, ensuring all of the localized fields get a value attached to them.


ryanmitchell:
Good find, thank you. Would you mind creating a test that shows the behaviour is fixed?

PR #479, opened 2025-07-24
[4.x] Always load model for assets

godismyjudge95:
This PR fixes the same issue I was trying (but failed) to address in this PR to core: https://github.com/statamic/cms/pull/11983

It solves the issue of assets being only partially loaded when using the `AssetContainer()->asset()` method.  That method is used by a few critical locations chief among them being the `GlideController` in core.  This is the method call chain that causes issue:
- `Statamic\Assets\AssetContainer->asset()`
- `Statamic\Assets\AssetContainer->makeAsset()`
- `Statamic\Assets\Asset->hydrate()`
- `Statamic\Eloquent\Assets\Asset->meta()`
- `Statamic\Eloquent\Assets\Asset->model()`

Notice that chain never calls the `Statamic\Eloquent\Assets\Asset->fromModel()` method like the `Statamic\Eloquent\Assets\AssetQueryBuilder` does.  This means that when `Asset->meta()` calls `Asset->model()` to get the meta from the current model it will be empty.

To solve this, I am simply adding a check that if the model is empty we will fill it using the result of a model query (copied from the `Statamic\Eloquent\Assets\Asset->makeModelFromContract()` method).

Let me know if there is anything else I can clarify or change.

### DIFF
--- a/src/Assets/Asset.php
+++ b/src/Assets/Asset.php
@@ -173,6 +173,14 @@ class Asset extends FileAsset
     public function model($model = null)
     {
         if (func_num_args() === 0) {
+            if (! $this->model) {
+                $this->model = app('statamic.eloquent.assets.model')::query()->where([
+                    'container' => $this->containerHandle(),
+                    'folder' => $this->folder(),
+                    'basename' => $this->basename(),
+                ])->first();
+            }
+
             return $this->model;
         }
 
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Thank you. I made a few changes to the tests mainly but its all good.

PR #494, opened 2025-08-27
Load model from config

kevinmeijer97:
When the sites driver is kept on file and this migration gets added to your project the following error occurs when running the migrations.

Screenshot 2025-08-27 at 13 42 28

This is correct because the model is only binded on the app if the sites driver is configured as eloquent.

Screenshot 2025-08-27 at 13 45 00

I changed the migration to use the model from the config instead, this way it's always available.

### DIFF
--- a/database/migrations/updates/add_order_to_sites_table.php.stub
+++ b/database/migrations/updates/add_order_to_sites_table.php.stub
@@ -13,7 +13,7 @@ return new class extends Migration {
         });
 
         $count = 0;
-        app('statamic.eloquent.sites.model')::all()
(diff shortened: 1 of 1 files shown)


ryanmitchell:
It would probably be better to update the UpdateScript for this migration to not publish if the site repository is not eloquent: https://github.com/statamic/eloquent-driver/blob/4.x/src/Updates/AddOrderToSitesTable.php

kevinmeijer97:
Agreed, changed it around to check for the config there instead!

ryanmitchell:
Looks good to me, thanks for making the changes.

PR #505, opened 2025-09-24
[5.x] Drop support for old asset container options

duncanmcclean:
We removed various options from Asset Containers in statamic/cms#12281, in favour of permissions. This pull request drops support for these options in the Eloquent Driver.


duncanmcclean:
Failing test is unrelated. Will fix shortly.

Issue #501, opened 2025-09-22
v4.31.0 sorting integers as strings


mthorsdal:
### Bug description

after upgrading from v4.30.0 to 4.31.0, my collection tag sorts integers as string

### How to reproduce

Configure collections and entries to use eloquent-driver,

Create a collection called "activities", blueprint activity.yaml with an integer field list_priority and add multiple entries to the database - with values such as 9 and 40, then in your antlers template, add

{{ collection:activities sort="list_priority:desc"}}
// I have also this filter, but I assume not relevant:  taxonomy:destinations="{destination_filter:slug}"

<h1>{{list_priority}}</h1>

{{ /collection:activities }}

And value 9 would appear before value 40

### Logs

```shell

```

### Environment

```yaml
Environment
Application Name: Pissup
Laravel Version: 11.46.0
PHP Version: 8.3.23
Composer Version: 2.5.8
Environment: local
Debug Mode: ENABLED
URL: pissup.test
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: redis
Database: mysql
Logs: stack / single
Mail: smtp
Queue: sync
Session: database

Livewire
Livewire: v3.6.4

Statamic
Addons: 10
Sites: 7 (Pissup.com, Pissup.de, Evg.fr, and 4 more)
Stache Watcher: Enabled
Static Caching: Disabled
Version: 5.65.1 PRO

Statamic Addons
aryehraber/statamic-logbook: 3.3.0
eminos/statamic-cloudflare-cache: dev-main
fdmind/ignore-query-strings: 3.0.0
jacksleight/statamic-bard-mutator: 3.0.5
joetannenbaum/antlers-stack-tags: 0.1.0
marcorieser/statamic-livewire: 4.5.3
ryanmitchell/statamic-translation-manager: 2.2.1
statamic/eloquent-driver: 4.31.0
stoffelio/statamic-404-logger: 1.1.0
withcandour/aardvark-seo: 5.0.5

Statamic Eloquent Driver
Asset Containers: file
Assets: eloquent
Blueprints: file
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: file
Forms: file
Global Sets: file
Global Variables: file
Navigation Trees: file
Navigations: file
Revisions: file
Sites: file
[...]

PR #513, opened 2025-10-22
Fix sorting by numbers when using collection `in` condition

jacksleight:
Fixes https://github.com/statamic/eloquent-driver/issues/501

I also adjusted the values in the test, as the original values didn't actually cause a failure when sorted as strings.


duncanmcclean:
Looks good to me.

PR #517, opened 2025-11-24
Fix outrageous query counts when using Eloquent driver and querying terms on entries

Michael-Stokoe:
## Problem
When using collection tags in Antlers templates, entries with taxonomy fields triggered individual database queries for each taxonomy term, causing major N+1 query performance issues.

## Solution
Implemented proper Eloquent relationships between entries and taxonomy terms using a pivot table, enabling eager loading to eliminate N+1 queries.

## Performance Impact
- **Before**: Actual query equation: 1 + (N entries × M taxonomy fields) queries
- **After**: 2 queries total
- **Example**: 10 entries with 2 taxonomy fields: 21 queries → 2 queries (90% reduction)

## Changes
- Added `entry_term` pivot table for entry-taxonomy relationships
- Enhanced Entry models with taxonomy relationships and auto-eager loading
- Added `SyncTaxonomyRelationships` command for existing data migration
- Maintains full backward compatibility - no template changes required

## Migration
```bash
php artisan vendor:publish --tag=statamic-eloquent-entry-term-pivot
php artisan migrate
php please eloquent:sync-taxonomy-relationships
```

## Backward Compatibility
✅ Existing templates work unchanged  
✅ API responses remain identical  
✅ No breaking changes

## Before and afters captured on a local site with real, live data

Before:
image

After:
image

---

Before:
CleanShot 2025-11-24 at 15 06 44@2x

After:
CleanShot 2025-11-24 at 15 07 10@2x


FrittenKeeZ:
We're also experiencing this issue, which has actually caused our comwell.com site to crash due to the database being overloaded, so this is the highest priority for us right now.

Michael-Stokoe:
The problem is that when navigation pages call setEntry(), it's loading entries which have the terms relationship eager loaded.

Those terms might themselves reference entries (if they have entry fields), creating a circular reference loop.

Pushing the fix for it now :)

PR #522, opened 2025-12-08
Fix taxonomy default blueprint where namespace is null

SteJW:
I reported a [bug](https://github.com/statamic/importer/issues/120) in the statamic importer, but it's an eloquent related problem. It's also causing the a problem in the overview of taxonomies (/cp/taxonomies) with the same error.
I created a pull request for a small fix.


SteJW:
Doesn't it still do that with blueprints that are not in the config, hence not in the database? This should only skips null namespaces, and look for them in the database. Would there be a way to have the null namespace look for the file repository?

PR #548, opened 2026-02-20
Fix and improve asset and collection export commands

sjongejan:
Fixed several issues and improved flexibility for data import and export commands.

*Import*
- Added `source_preset` to the asset container export

*Export*
- Added support for `--only-asset-containers` and `--only-assets` flags in the assets export command
- Added support for `--only-collections` and `--only-collection-trees` flags in the collection export command
- Added support for `--only-navs` and `--only-nav-trees` flags in the collection export command
- Fixed bug in assets export
- Enabled independent exporting of collection trees. This allows collections to be maintained in static files while their trees are stored in the database (Eloquent).


ryanmitchell:
Thanks for the PR and the work on it, all looks great.

PR #569, opened 2026-04-17
Use saveQuietly() in ExportEntries to prevent infinite loop

### LINKED ISSUE #567: eloquent:export-entries hangs indefinitely on structured collections with propagate: true
> **Note:** This issue was created and analyzed with AI assistance (Claude Code).

## Bug description

Running `php please eloquent:export-entries` causes an infinite CPU loop when exporting entries from a collection that has both `structure: root: true` and `propagate: true` (e.g. a typical Pages collection in a multi-site setup).

The command internally swaps repositories to Stache and calls `$entry->save()`. For structured collections with propagation enabled, this triggers a chain of tree rebuilds and propagation events that never terminates. The process spins at 100% CPU and never progresses past the first structured entry.

Using `$entry->saveQuietly()` instead of `$entry->save()` on lines 113 and 144 of `ExportEntries.php` resolves the issue completely — all entries export in seconds. Since the command is only writing flat files, firing events during export is unnecessary and counterproductive.

Even after the entries are written, the process hangs on shutdown due to Stache index finalization hitting the same loop.

## How to reproduce

1. Create a multi-site Statamic app with the Eloquent driver for entries
2. Have a collection with `structure: root: true` and `propagate: true` (default Pages collection)
3. Run `php please eloquent:export-entries`
4. Observe the progress bar stopping and CPU usage going to 100%

## Expected behavior

All entries should export to flat files without hanging.

## Environment

- Laravel: v13.5.0
- Statamic: v6.13.0
- Eloquent Driver: v5.6.2
- PHP: 8.5.4
- Database: PostgreSQL
- Sites: 6

lunitrixx:
## Summary

- Replaces `$entry->save()` with `$entry->saveQuietly()` in `ExportEntries` command
- Prevents an infinite loop of tree rebuilds and propagation events when exporting structured collections with `propagate: true`
- Since the command only writes flat files, firing events during export is unnecessary

Fixes #567

### DIFF
--- a/src/Commands/ExportEntries.php
+++ b/src/Commands/ExportEntries.php
@@ -99,7 +99,7 @@ class ExportEntries extends Command
                 $entry->set('updated_at', $model->updated_at ?? $model->created_at);
             }
 
-            $entry->save();
+            $entry->saveQuietly();
         });
 
         if ($entriesWithOrigin->count() > 0) {
@@ -130,7 +130,7 @@ class ExportEntries extends Command
                     $entry->set('updated_at', $model->updated_at ?? $model->created_at);
                 }
 
-                $entry->save();
+                $entry->saveQuietly();
             });
         }
 


ryanmitchell:
thank you

PR #577, opened 2026-04-30
Fix Submission::date() to fall back to id-derived timestamp

### LINKED ISSUE #576: Eloquent Submission::date() breaks parent invariant — falls back to now() instead of id-derived timestamp
### Bug description

`Statamic\Eloquent\Forms\Submission` extends `Statamic\Forms\Submission` but overrides `date()` in a way that breaks the parent's invariant:
                                                                                                                                                                                                
  > A submission with a numeric-timestamp `id` (Statamic's `{unix-ts}.{micro}.yaml`              
  > filename convention) MUST yield a `date()` derived from that id when no                                                                                                                                                                              
  > explicit date has been set.                          
                                                                                                                                                                                                                                                         
Parent ([vendor/statamic/cms/src/Forms/Submission.php:112-115](https://github.com/statamic/cms/blob/6.x/src/Forms/Submission.php#L112-L115)):
```php
public function date()
{
    return Carbon::createFromTimestamp($this->id());
}
```

Eloquent child ([src/Forms/Submission.php:58-71](https://github.com/statamic/eloquent-driver/blob/5.x/src/Forms/Submission.php#L58-L71)):
```php
public function date($date = null)
{
    if (! is_null($date)) { ... }
    return $this->date ?? ($this->date = Carbon::now());  // this breaks invariant
}                                                                                              
```
                                                                                                                                                                                                                                                         
Result: any code path that obtains an Eloquent-child Submission whose `$date` hasn't been hydrated, then asks for `date()`, gets `now()`.


### How to reproduce

This happens when exec `statamic:eloquent:import-forms` in certain conditions. Check `saving_uses_id_derived_timestamp_when_date_is_unset()` regression test without the fix.

### Logs

```shell

```

### Environment

```yaml
Laravel Version: 12.58.0
PHP Version: 8.4.20
Environment: development
Debug Mode: ENABLED

**Cache**
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

**Statamic Addons**
statamic-rad-pack/runway: 9.4.4
statamic/eloquent-driver: 5.6.3
```

### Additional details

_No response_

SUXUMI:
Fixes statamic/eloquent-driver#576

### DIFF
--- a/src/Forms/Submission.php
+++ b/src/Forms/Submission.php
@@ -67,7 +67,15 @@ class Submission extends FileEntry
             return $this;
         }
 
-        return $this->date ?? ($this->date = Carbon::now());
+        if ($this->date) {
+            return $this->date;
+        }
+
+        if (is_numeric($this->id)) {
+            return $this->date = Carbon::createFromTimestamp($this->id);
+        }
+
+        return $this->date = Carbon::now();
     }
 
     public function save()
--- a/tests/Forms/FormSubmissionTest.php
+++ b/tests/Forms/FormSubmissionTest.php
@@ -103,7 +103,7 @@ class FormSubmissionTest extends TestCase
         $this->assertInstanceOf(Carbon::class, $submission->date());
         $this->assertArrayNotHasKey('date', $submission->model()->data);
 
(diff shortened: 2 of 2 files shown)


ryanmitchell:
Thank you.

Issue #578, opened 2026-05-05
TermQueryBuilder::where('site', '=', $handle) silently stores `=` as the site (3-arg form bypasses normalization)


SUXUMI:
### Bug description

`Statamic\Eloquent\Taxonomies\TermQueryBuilder::where()` special-cases the `site` column *before* the standard 2-arg → 3-arg operator normalization runs. Calling it with the standard Laravel 3-arg form `where('site', '=', $value)` stores the **operator** (`'='`) as the site, not the value.

 This is silent: no exception, no warning. `\Statamic\Taxonomies\LocalizedTerm::in('=')` then produces a term locked to a non-existent site, and value lookups cascade through to defaults or return nothing.

**Source:** `src/Taxonomies/TermQueryBuilder.php:57-67`

 ```php
 public function where($column, $operator = null, $value = null, $boolean = 'and')
 {
     if ($column === 'site') {
         $this->site = $operator;   // <-- captures whatever 2nd arg is
         return $this;
     }

     if (func_num_args() === 2) {
         [$value, $operator] = [$operator, '='];   // normalization runs AFTER
     }
     // ...
 }
```

### Environment

```yaml
- statamic/eloquent-driver: 5.6.3
- PHP 8.4
```

ryanmitchell:
We intentionally don't support comparison on the site column - the assumption is you will always be querying `where('site', 'site-handle')`. If you look around the TermQueryBuilder you see there are places we reference the site to return the right localised term.

its possible to support this, but we'd need to do a larger piece of work to change how term data is stored.

SUXUMI:
Sorry, I missed that. My apologies.

ryanmitchell:
All good, Ive worked up the [bigger change here](https://github.com/statamic/eloquent-driver/pull/580), but its likely breaking so will have to wait for the next major release.

PR #579, opened 2026-05-05
Fix TermQueryBuilder site filter to support 3-arg where() form

SUXUMI:
#578


ryanmitchell:
Unfortunately as I detailed in my reply this isn't a sufficient change. Let me provide a worked example:

If you have an 'en' site that has a localisation in 'fr', then the localisation is stored in the data->localisations on the 'en' term.

If you tried to query for 'fr' terms this term would not be returned.

Additionally, if you tried to query ->where('site', '!=', 'handle') this PR would still treat it like where('site', '=', 'handle')

SUXUMI:
Sorry, missed that..

PR #580, opened 2026-05-05
[6.x] Store term localizations as separate rows 

### LINKED ISSUE #578: TermQueryBuilder::where('site', '=', $handle) silently stores `=` as the site (3-arg form bypasses normalization)

ryanmitchell:
Previously, all locale data for a term was embedded in a single database row under a data->localizations JSON key. This made it impossible to query terms by site using a real WHERE site = ? clause, since translations stored inside a row's JSON are invisible to SQL.                                                         
                                                                                                                                                                           
This PR moves each localization to its own database row. The canonical (default-locale) row keeps origin = NULL, and each translation row stores origin = <canonical_id>.

This aligns with how entries already work.

Closes #578

### DIFF
--- a/database/migrations/2025_07_07_200000_add_origin_to_taxonomy_terms.php
+++ b/database/migrations/2025_07_07_200000_add_origin_to_taxonomy_terms.php
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Statamic\Eloquent\Database\BaseMigration as Migration;
+
+return new class extends Migration
+{
+    public function up()
+    {
+        Schema::table($this->prefix('taxonomy_terms'), function (Blueprint $table) {
+            $table->unsignedBigInteger('origin')->nullable()->after('site');
+            $table->index('origin');
+        });
+    }
+
+    public function down()
+    {
+        Schema::table($this->prefix('taxonomy_terms'), function (Blueprint $table) {
+            $table->dropIndex(['origin']);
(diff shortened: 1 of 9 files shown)


ryanmitchell:
@duncanmcclean this is likely a breaking change given the underlying storage changes. what do you think?

duncanmcclean:
I like the idea. It would certainly make localizations easier to query.

However, as you say, it's likely a breaking change so it'll probably need to wait until the next major release, whenever that is (probably Q1, alongside v7) 😞.

PR #583, opened 2026-05-05
Change revisions message column to be text on new installs

### LINKED ISSUE #582: Revision import fails when YAML revision message exceeds 255 characters
### Bug description

`php please eloquent:import-revisions` fails when a flat-file revision YAML contains a `message` longer than 255 characters (for strict db connection).

The Eloquent driver migration creates `revisions.message` as `string`, which becomes `VARCHAR(255)` on MySQL: `database/migrations/2024_03_07_100000_create_revisions_table.php`

The import command passes the YAML message through unchanged and                                            
[src/Revisions/Revision.php::toModel()](https://github.com/statamic/eloquent-driver/blob/5.x/src/Commands/ImportRevisions.php#L57) then persists it as-is.

### Environment

```yaml
- statamic/eloquent-driver: 5.6.3
- PHP 8.4
```

ryanmitchell:
Closes #582 

@duncanmcclean My feeling is we don't need to back port this to existing installs. If someone comes against it they can make a migration... what do you think?

### DIFF
--- a/database/migrations/2024_03_07_100000_create_revisions_table.php
+++ b/database/migrations/2024_03_07_100000_create_revisions_table.php
@@ -13,7 +13,7 @@ return new class extends Migration
             $table->string('key');
             $table->string('action')->index();
             $table->string('user')->nullable();
-            $table->string('message')->nullable();
+            $table->text('message')->nullable();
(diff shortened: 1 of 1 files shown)


duncanmcclean:
Agreed.

PR #586, opened 2026-05-07
Forget the all-list Blink cache when saving containers, taxonomies and globals

### LINKED ISSUE #585: Repository::save() leaves the all-list Blink cache stale, so reads after writes return pre-save state in the same request
### Bug description

After `AssetContainerRepository::save()` persists a new container, the cached result of `AssetContainer::all()` (Blink key `eloquent-assetcontainers`) is not invalidated. Any subsequent `all()` call within the same request returns the pre-save list, so the new container is invisible to the rest of the request.

`delete()` correctly forgets both the per-handle key and the all-list key. `save()` only forgets the per-handle key.

The same shape exists in two sibling repositories:

- `TaxonomyRepository::save()` calls `Blink::put` per-handle but never forgets `eloquent-taxonomies`.
- `GlobalRepository::save()` forgets the per-handle key but never `eloquent-globalsets`.

`CollectionRepository` is the only repo that gets it right and serves as the reference: it forgets both keys on save *and* delete.

This is the write-path mirror of #528 / #533, where `findByHandle` was caching null on miss. Same Blink invalidation oversight, opposite direction.

### How to reproduce

Three lines in tinker against any project where `asset_containers` uses the eloquent driver:

```php
use Statamic\Facades\{AssetContainer, Blink};

Blink::forget('eloquent-assetcontainers');
AssetContainer::all()->count();                       // caches the list = N
AssetContainer::make('test')->disk('local')->save();  // DB updated, Blink not
AssetContainer::all()->count();                       // returns N, missing 'test' (BUG)
```

Real-world surface where this bites: a Laravel seeder that creates the `assets` container and then asks for it via `AssetContainer::findByHandle('assets')` in the same artisan process. The lookup misses because the cached empty list is consulted before the DB, so the seeder silently no-ops or `Asset::make()->container('assets')->save()` later crashes on a null container handle.

Source pointers (5.x):

- `src/Assets/AssetContainerRepository.php:44` — `save()` forgets only `eloquent-assetcontainer-{handle}`
- `src/Assets/AssetContainerRepository.php:51` — `delete()` forgets both, ✅ reference
- `src/Taxonomies/TaxonomyRepository.php:85` — same shape (uses `Blink::put` per-handle, never forgets `eloquent-taxonomies`)
- `src/Globals/GlobalRepository.php:45` — same shape (forgets per-handle only)
- `src/Collections/CollectionRepository.php:33` — ✅ correct, forgets both

### Logs

_No errors logged; the bug is silent. Failing assertion in tinker:_

```text
> AssetContainer::all()->count()
= 1
> AssetContainer::make('new')->disk('local')->save()
= Statamic\Eloquent\Assets\AssetContainer {…}
> AssetContainer::all()->count()
= 1   // expected 2
```

### Environment

```yaml
Environment
Laravel Version: 13.2.0
PHP Version: 8.4.20
Composer Version: 2.9.7

Statamic
Version: 6.12.0 PRO

Statamic Addons
statamic/eloquent-driver: 5.6.2 (also reproduced against tip of 5.x and v5.7.0)

Statamic Eloquent Driver per-domain config in the reproducing project
Asset Containers: eloquent
Taxonomies: eloquent
Global Sets: file (so the GlobalRepository case is source-only,
                   not live-reproduced; the eloquent path is the
                   one this issue covers)
```

### Additional details

I'm happy to send a PR. It's a one-line addition per affected `save()` method, mirroring what each `delete()` already does:

```php
// AssetContainerRepository::save
Blink::forget('eloquent-assetcontainers');

// TaxonomyRepository::save
Blink::forget('eloquent-taxonomies');

// GlobalRepository::save
Blink::forget('eloquent-globalsets');
```

I have a branch ready with the fix and a regression test in each `*RepositoryTest`. The tests fail on tip of `5.x` and pass with the fix. Let me know if you'd prefer separate issues per repository.

BartWaardenburg:
## Summary

- Add `Blink::forget('eloquent-assetcontainers')` to `AssetContainerRepository::save()`
- Add `Blink::forget('eloquent-taxonomies')` to `TaxonomyRepository::save()`
- Add `Blink::forget('eloquent-globalsets')` to `GlobalRepository::save()`
- Add a regression test per repository

Mirrors what each `delete()` already does so reads after writes within the same request reflect the new entry rather than the pre-save cached list. `CollectionRepository` was the only repo that got this right; the others now match.

The tests fail on upstream `5.x` and pass with this change. Drafted while triage on #585 is open. Happy to scope down to one repository if you'd prefer separate PRs.

Fixes #585

### DIFF
--- a/src/Assets/AssetContainerRepository.php
+++ b/src/Assets/AssetContainerRepository.php
@@ -46,6 +46,7 @@ class AssetContainerRepository extends StacheRepository
         $container->save();
 
         Blink::forget("eloquent-assetcontainer-{$container->handle()}");
+        Blink::forget('eloquent-assetcontainers');
     }
 
     public function delete($container)
--- a/src/Globals/GlobalRepository.php
+++ b/src/Globals/GlobalRepository.php
@@ -54,6 +54,7 @@ class GlobalRepository extends StacheRepository
         $entry->model($model->fresh());
 
         Blink::forget("eloquent-globalsets-{$model->handle}");
+        Blink::forget('eloquent-globalsets');
     }
 
     public function delete($entry)
--- a/src/Taxonomies/TaxonomyRepository.php
+++ b/src/Taxonomies/TaxonomyRepository.php
@@ -92,6 +92,7 @@ class TaxonomyRepository extends StacheRepository
         $entry->model($fresh);
 
         Blink::put("eloquent-taxonomies-{$fresh->handle}", $fresh);
+        Blink::forget('eloquent-taxonomies');
     }
 
     public function delete($entry)
--- a/tests/Repositories/AssetContainerRepositoryTest.php
+++ b/tests/Repositories/AssetContainerRepositoryTest.php
@@ -73,4 +73,17 @@ class AssetContainerRepositoryTest extends TestCase
         $this->assertNotNull($item = $this->repo->findByHandle('new'));
         $this->assertEquals($container, $item);
     }
+
+    #[Test]
+    public function saving_a_container_invalidates_the_cached_all_list()
+    {
+        // Warm the Blink cache by reading all() before the new container exists.
+        $this->assertCount(2, $this->repo->all());
+
+        $this->repo->save(Facades\AssetContainer::make('new')->disk('local'));
+
+        $handles = $this->repo->all()->map->handle()->all();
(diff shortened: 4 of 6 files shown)


ryanmitchell:
Great spot, thanks for this.

PR #587, opened 2026-05-13
[4.x] Use closest origin that has field localized to determine if a field can be forgotten in `makeModelFromContract` method of `Entry`

Boefjim:
When working with localizations of localizations, the localized fields are not handled appropriately, specifically whenever a localization field is left empty it only checks the direct origin and sees null == null and thus it think it does not need to be localized (even though the origin that the value is then inherited from might not be null).

This change will make it so for each localized field, it will check the closest origin that has the field localized, ensuring that this is cascaded correctly.

While I personally need this fixed in the 4.x branch, it seems this issue also exists in the 5.x branch...should I make a separate PR for 5.x or how would it be best to approach this?
The method is slightly expanded in the new version, but the logic that determines which localized fields can be forgotten seems to be the same, and could thus likely receive the same patch.

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -61,7 +61,7 @@ class Entry extends FileEntry
 
         $date = $source->hasDate() ? $source->date() : null;
 
-        $origin = $source->origin();
+        $directOrigin = $source->origin();
 
         if ($source->hasOrigin()) {
             if ($blueprint = $source->blueprint()) {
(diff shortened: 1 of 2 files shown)


ryanmitchell:
Would you please add a test showing that this change fixes the incorrect behaviour?

Boefjim:
> Would you please add a test showing that this change fixes the incorrect behaviour?

Yes, I will try to get a test written for this, although it might take some time, since I have only managed to reproduce this through the Statamic UI and I am unsure what methods are called through it in order to reproduce this through code 😅

Boefjim:
@ryanmitchell I have added a test that should reproduce the issue I was experiencing.
I have also put the commit that adds the test before the fix, so it is easier to see it failing by simply checking out that commit (hence the force-push).

Let me know if you need any additional changes / information!

ryanmitchell:
Thank you

PR #598, opened 2026-06-03
Use closest origin that has field localized

ryanmitchell:
Duplicate of https://github.com/statamic/eloquent-driver/pull/587 but for 5.x branch

### DIFF
--- a/src/Entries/Entry.php
+++ b/src/Entries/Entry.php
@@ -62,7 +62,7 @@ class Entry extends FileEntry
 
         $date = $source->hasDate() ? $source->date() : null;
 
-        $origin = $source->origin();
+        $directOrigin = $source->origin();
 
         if ($template = $source->get('template', $source->template)) {
             $data->put('template', $template);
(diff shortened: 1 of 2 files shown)


src/Entries/Entry.php
```diff
@@ -82,26 +82,36 @@ public static function makeModelFromContract(EntryContract $source)
                     $data->forget('template');
                 }
 
-                $originData = $origin->data();
+                $directOriginData = $directOrigin->data();
 
-                // remove any fields in entry data that are marked as localized but value is present, and matches origin
+                // remove any fields in entry data that are marked as localized but value is present, and does not match origin
```
Boefjim, line 87: @ryanmitchell 

I think this reverted a fix for this comment on the 5.x branch, that had not been applied on the 4.x branch?

This should only remove fields that are present & match with the origin (or now rather the closest origin that localizes the field), so the original comment here seems to be more accurate 🤔

ryanmitchell: Thanks, I've updated that directly on 5.x

PR #604, opened 2026-08-10
Preserve the order of bard and replicator sets

### LINKED ISSUE #15152: Order of Bard Sets Not Saved
### Bug description

When I try to reorder the available sets in a Bard field, the ordering is not saved.

I'm using eloquent and blueprints are saved in the database.

### How to reproduce

Create a blueprint with a Bard field and multiple sets. Try reordering the sets and see if the ordering is preserved when adding a set in an entry.

### Logs

```shell

```

### Environment

```yaml
Environment
Laravel Version: 12.63.0
PHP Version: 8.3.22
Composer Version: 2.8.9
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: en

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: postmark
Queue: sync
Session: database

Storage
public/storage: LINKED

Livewire
Livewire: v3.7.10

Statamic
Addons: 1
License Key: Set
Stache Watcher: Enabled
Static Caching: Disabled
Version: 6.27.1 PRO

Statamic Addons
statamic/eloquent-driver: 5.10.0

Statamic Eloquent Driver
Addon Settings: file
Asset Containers: file
Assets: file
Blueprints: eloquent
Collection Trees: file
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: file
Forms: eloquent
Global Sets: eloquent
Global Variables: file
Navigation Trees: file
Navigations: eloquent
Revisions: eloquent
Sites: file
Taxonomies: eloquent
Terms: eloquent
Tokens: file
```

### Installation

Existing Laravel app

### Additional details

_No response_

lazerg:
Reordering the sets on a Bard or Replicator field doesn't stick when blueprints are stored with the eloquent driver on MySQL. MySQL sorts the keys of a JSON object when it stores it, so the sets come back out in a different order to the one they went in with.

The blueprint repository already works around this for tabs and sections by keeping a `__count` on each of them. This does the same for the set groups and sets nested in a field's config, and strips it back out when the blueprint is read or exported. The import command gets it too, since it writes to the database without going through the repository.

Fieldsets have the same problem and aren't covered here, as `FieldsetRepository` does no order handling at all and would need its own pass. Happy to follow up with that if you want it.

Fixes statamic/cms#15152.


tests/Data/Fields/BlueprintTest.php
```diff
@@ -81,49 +82,33 @@ public function it_preserves_the_order_of_sets()
...
-                                                ],
-                                            ],
-                                        ],
-                                    ],
-                                ],
-                            ],
-                        ],
-                    ],
-                ],
+                'tabs' => ['main' => ['sections' => [['fields' => [['handle' => 'content', 'field' => [
```
ryanmitchell: please undo the formatting change

lazerg: Reverted, sorry about that. It crept in with an unrelated tidy up.

ryanmitchell:
I've applied it to fieldsets too for completeness. thanks for your work on this.

PR #605, opened 2026-08-10
[6.x] Match entries with an unset field when querying with a not equals operator

### LINKED ISSUE #603: Entries with an unset field are matched by the Stache but not by the eloquent driver
### Bug description

When a field is absent from an entry's stored data, the two drivers disagree about whether that entry matches a comparison against the field.

For example, I have a toggle in Statamic for "archived" that was added to the blueprint after the collection and entries already existed.

The Stache treats the missing field as falsy, so the entry matches where('field', '!=', true) and where('field', false). The eloquent driver compares json_extract(data, '$.field'), which is NULL, and SQL comparisons against NULL are never true — so the entry matches neither.

The practical effect is that a template like {{ collection:things archived:isnt="true" }} renders every entry under the file driver and nothing under the eloquent driver, as soon as the entries in question never had that toggle switched on and off.

> [!NOTE]
> Disclaimer, the bug is real and causing my tests to fail in different environments... I use eloquent-driver on production but not on localhost. Claude helped me write the Github Issue, sorry about that.

### How to reproduce

A collection where no entry has ever had the archived toggle set, so the key is absent from data for all 48 entries:

```php
  use Statamic\Facades\Entry;

  $q = fn () => Entry::query()->where('collection', 'volumes')->where('site', 'en');

  $q()->count();                                 // file: 48   eloquent: 48
  $q()->where('archived', '!=', true)->count();  // file: 48   eloquent: 0   ❌
  $q()->where('archived', false)->count();       // file: 48   eloquent: 0   ❌
  $q()->whereNull('archived')->count();          // file: 48   eloquent: 48  ✅
```

Same content, same query, switched only by STATAMIC_ENTRIES_DRIVER.

Reached from a template, this is simply:

  ```antlers
  {{ collection:volumes archived:isnt="true" }}
```

Statamic\Tags\Parameters casts the "true" string to a real boolean, so the tag ends up calling where('archived', '!=', true) — the first failing case above.

###  Expected
  
An entry whose toggle was never switched on should be treated as not-true, matching the Stache. An unset toggle and an explicitly-false toggle look the same to an editor, and the file driver treats them the same way; the eloquent driver does not.

###  Why it is easy to miss

Whether it breaks depends on whether the key happens to exist in your data, which depends on the editing history of each entry rather than anything visible in the blueprint. In our app, one collection had the key on every entry and behaved correctly, while another had it on none and rendered an empty page. Both used the same template pattern.

There is no error, no log line, and no failed query — the listing is simply empty.

###  Workaround

Being explicit about the null case gives identical results under both drivers:                                                                 

```php
  // instead of ->where($field, '!=', true)
  $query->where(fn ($q) => $q->whereNull($field)->orWhere($field, false));
```

We moved the affected filters out of tag conditions into a query scope that does this.

###  Possible fix

Comparisons against a JSON key that may be absent could be compiled to include the NULL case — e.g. treating where($field, '!=', $value) as  json_extract(...) IS NULL OR json_extract(...) != $value when the field is a data key rather than a real column. That would match the Stache's  behaviour of treating a missing key as null/falsy.

### Logs

```shell

```

### Environment

```yaml
- statamic/cms v6.20.0
  - statamic/eloquent-driver v5.9.0
  - Laravel 13.12.0, PHP 8.4.22
  - SQLite, entries stored with data as a JSON column
```

### Additional details

Only verified on SQLite, though the NULL comparison semantics should be the same on MySQL and Postgres.

lazerg:
When a blueprint field has never been set on an entry, its key is simply missing from the data column, so json_extract returns NULL and `where('archived', '!=', true)` skips the entry. The Stache matches it. A template like `{{ collection:things archived:isnt="true" }}` therefore lists everything on the file driver and nothing here.

I've deliberately kept this much narrower than the change suggested in the issue. It only applies to `!=` and `<>`, and only to a top level field key in the data column. Real columns are left alone, so `where('date', '!=', ...)` still follows normal SQL null rules. Nested paths are untouched too, so `where('content->value', '<>', 1)` still ignores entries whose content isn't shaped that way, as `entries_are_found_using_where_with_json_value` expects. No existing test needed changing.

Fixes #603.

### DIFF
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -115,10 +115,15 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
             trigger_error('Filtering by status is deprecated. Use whereStatus() instead.', E_USER_DEPRECATED);
         }
 
+        $actualColumn = $this->column($column);
+
+        // A field that was never set isn't in the data column at all, so json_extract gives
+        // us null and the entry falls out of the results. The Stache matches it, so we do too.
         if (
             func_num_args() > 2
+            && ! is_null($value)
             && in_array($operator, ['!=', '<>'], true)
-            && $this->isFieldColumn($actualColumn = $this->column($column))
+            && $this->isTopLevelDataColumn($actualColumn)
         ) {
             $this->builder->where(function ($query) use ($actualColumn, $operator, $value) {
                 $query->whereNull($actualColumn)->orWhere($actualColumn, $operator, $value);
@@ -139,7 +144,7 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
         return parent::whereIn(...func_get_args());
     }
 
-    private function isFieldColumn($column)
+    private function isTopLevelDataColumn($column)
     {
         return Str::contains($column, 'data->') && ! Str::contains(Str::after($column, 'data->'), '->');
     }
--- a/src/Entries/EntryQueryBuilder.php
+++ b/src/Entries/EntryQueryBuilder.php
@@ -115,6 +115,18 @@ class EntryQueryBuilder extends EloquentQueryBuilder implements QueryBuilder
             trigger_error('Filtering by status is deprecated. Use whereStatus() instead.', E_USER_DEPRECATED);
         }
 
+        if (
+            func_num_args() > 2
+            && in_array($operator, ['!=', '<>'], true)
+            && $this->isFieldColumn($actualColumn = $this->column($column))
+        ) {
+            $this->builder->where(function ($query) use ($actualColumn, $operator, $value) {
+                $query->whereNull($actualColumn)->orWhere($actualColumn, $operator, $value);
+            }, null, null, $boolean);
+
+            return $this;
+        }
+
         return parent::where(...func_get_args());
     }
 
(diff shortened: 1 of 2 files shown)


ryanmitchell:
As I said in the issue, I'm still not convinced this is necessary. With eloquent you can make a migration to backfill the data, and most developers using eloquent driver would understand and expect to do that.

This PR basically forces a whereNull check for any field you query, which is not necessaroly what you would want if you are doing direct queries.

Happy to be overruled @duncanmcclean

lazerg:
One clarification on the scope, since "any field you query" is broader than what this actually does.

The rewrite only applies to `!=` and `<>`, and only when the column resolves to a top level key in the data array. Real database columns never reach it, so direct queries against those keep normal SQL null behaviour. Nested paths don't reach it either, so your `content->value` test from 2022 is unchanged and still passes. The test file is additions only.

The rest is your call, I'm happy either way.

ryanmitchell:
Yes I understand the scope.

However it would mean if I run `$query->where('my_field', '<>', 1)` it will now actually run `$query->where('my_field', '<>', 1)->orWhereNull('my_field')` which I don't think its correct, even if it matches the stache.

ryanmitchell:
Chatted with Duncan off-thread on this and as its a breaking change we'll target the next major version with this and document it in the upgrade guide.

lazerg:
Sounds good, thanks for checking with Duncan and finding a home for this in 6.x. Happy to wait for the next major version, and appreciate you covering it in the upgrade guide.

lazerg:
Closing this since it has been inactive for a while.

ryanmitchell:
Its going to be included in the next major version so reopening

PR #607, opened 2026-08-14
Hierarchical Taxonomy Support

ryanmitchell:
Complements https://github.com/statamic/cms/pull/15192


duncanmcclean:
That was quick!

PR #608, opened 2026-08-25
Add TokenRepository::all() for unexpired tokens

jackmcdade:
## Summary
- Implements `TokenRepository::all()` so this package satisfies the new method on `Statamic\Contracts\Tokens\TokenRepository` (statamic/cms#15274).
- Returns unexpired tokens only, matching the file repository.

## Test plan
- [ ] `TokenRepositoryTest::it_gets_all_non_expired_tokens` passes
- [ ] Sharing a draft preview link twice on an Eloquent-driver site reuses the same token

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Tokens/TokenRepository.php
+++ b/src/Tokens/TokenRepository.php
@@ -2,11 +2,21 @@
 
 namespace Statamic\Eloquent\Tokens;
 
+use Illuminate\Support\Collection;
 use Statamic\Contracts\Tokens\Token as TokenContract;
 use Statamic\Tokens\TokenRepository as Repository;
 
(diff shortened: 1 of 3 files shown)


ryanmitchell (commented):
Shouldn't all() just return all regardless of whether they are expired or not?

jackmcdade:
> Shouldn't all() just return all regardless of whether they are expired or not?

Yup. Silly robot. It got a bit eager with this PR.

duncanmcclean:
Marking as a draft until the Core PR is merged.

PR #610, opened 2026-08-31
Decode percent-encoded paths in `AssetRepository::findByUrl()`

### LINKED ISSUE #609: findByUrl() does not decode percent-encoded paths, returning null for filenames with spaces, parentheses or @
### Bug description

`Statamic\Eloquent\Assets\AssetRepository::findByUrl()` looks up the asset path without decoding it first, unlike the core repository it replaces. Asset URLs are percent-encoded, so any asset whose filename contains a character that gets encoded (spaces, parentheses, `@`, accented characters, commas) cannot be resolved from its own URL and returns `null`.

**Expected:** `Asset::find($asset->url())` returns the asset it was given.
**Actual:** it returns `null` for any filename that needs URL encoding.

The cause is a single missing call. Core (`statamic/cms`, `src/Assets/AssetRepository.php:62`):

```php
$path = rawurldecode(Str::after($url, $containerUrl));
```

Eloquent driver (`src/Assets/AssetRepository.php:56`):

```php
$path = Str::after($url, $containerUrl);
```

Without the decode, `findById()` is handed `my%20file%20%281%29.jpg` and queries the database for that literal `basename`, which matches no row.

In our production install this made every page carrying such an asset as its Open Graph image return a 500. See "Additional details" for why a failed lookup escalates that far.

### How to reproduce

1. Enable the eloquent driver for assets.
2. Upload an asset with a space, parenthesis or `@` in its filename, e.g. `my file (1).jpg`.
3. Resolve that asset from its own URL:

```php
$asset = Asset::find('assets::my file (1).jpg');

$asset->url();                      // "/assets/my%20file%20%281%29.jpg"
Asset::find($asset->url());         // null  <- expected the same asset back
```

Any filename that survives a round trip through `url()` unencoded works fine, which is why this stays hidden until someone uploads a file with a space in it.

To see it take down a page, put such an asset in SEO Pro's Open Graph image field and request any page that uses it.

### Logs

```shell
Call to a member function isVideo() on null
Error: Call to a member function isVideo() on null at /var/www/html/vendor/statamic/cms/src/Imaging/ImageGenerator.php:158

#0 /var/www/html/vendor/statamic/cms/src/Tags/Glide.php(180): Statamic\Imaging\ImageGenerator->generateByAsset()
#1 /var/www/html/vendor/statamic/cms/src/Tags/Glide.php(139): Statamic\Tags\Glide->generateImage()
#2 [internal function]: Statamic\Tags\Glide->{closure:Statamic\Tags\Glide::generate():134}()
#5 /var/www/html/vendor/statamic/cms/src/Tags/Glide.php(134): Illuminate\Support\Collection->map()
#6 [internal function]: Statamic\Tags\Glide->generate()
#17 /var/www/html/vendor/statamic/seo-pro/src/RendersMetaHtml.php(17): Illuminate\View\View->render()
#18 /var/www/html/vendor/statamic/seo-pro/src/Tags/SeoProTags.php(31): Statamic\SeoPro\Tags\SeoProTags->renderMetaHtml()
#30 /var/www/html/vendor/statamic/cms/src/Http/Responses/DataResponse.php(155): Statamic\View\View->render()
#32 /var/www/html/vendor/statamic/cms/src/Entries/Entry.php(614): Statamic\Http\Responses\DataResponse->toResponse()
```

### Environment

```yaml
Environment
Laravel Version: 12.68.0
PHP Version: 8.5.2
Composer Version: 2.10.2
Environment: local
Debug Mode: ENABLED
Maintenance Mode: OFF
Timezone: UTC
Locale: nl

Cache
Config: NOT CACHED
Events: NOT CACHED
Routes: NOT CACHED
Views: CACHED

Drivers
Broadcasting: log
Cache: file
Database: mysql
Logs: stack / single
Mail: log
Queue: sync
Session: file

Storage
public/storage: LINKED

Statamic
Addons: 7
License Key: Not set
Sites: 19
Stache Watcher: Enabled (auto)
Static Caching: Disabled
Version: 6.29.0 PRO

Statamic Addons
appswithlove/statamic-one-click-content-translation: 6.0.6
eminos/statamic-tabs: 2.0.1
keyagency/content-importer: dev-main
keyagency/kai-personalize: 1.2.12
keyagency/statamic-asset-usage: 1.0.1
statamic/eloquent-driver: 5.11.0
statamic/seo-pro: 7.13.3

Statamic Eloquent Driver
Addon Settings: eloquent
Asset Containers: file
Assets: eloquent
Blueprints: eloquent
Collection Trees: eloquent
Collections: eloquent
Entries: eloquent
Fieldsets: file
Form Submissions: eloquent
Forms: eloquent
Global Sets: eloquent
Global Variables: eloquent
Navigation Trees: eloquent
Navigations: eloquent
Revisions: eloquent
Sites: eloquent
Taxonomies: eloquent
Terms: eloquent
Tokens: eloquent
```

### Additional details

`Statamic\Tags\Glide::generateImage()` passes the result straight through:

```php
return $this->getGenerator()->generateByAsset(Asset::find($item), $params);
```

`ImageGenerator::generateByAsset()` has no null check, so line 158 throws `Error: Call to a member function isVideo() on null`. Because that is an `\Error` and `Glide::generate()` only catches `\Exception`, it escapes the tag's own error handling and takes down the entire response instead of just skipping the image.

The path that reaches this in practice is an `Asset` instance passed to the glide tag: `normalizeItem()` returns it unchanged, and `Asset::find()` then casts it to a string via `Asset::__toString()`, which yields the encoded `url()`. SEO Pro's `meta.antlers.html` does exactly this for `og:image`, so a single asset with a space in its name 500s every page that uses it. Static caching never stores the error response, so it recurs on every request.

The missing null check in `ImageGenerator` is arguably a second bug, but it lives in `statamic/cms` rather than in this repository.

### Suggested fix

Restore the decode to match core:

```diff
-        $path = Str::after($url, $containerUrl);
+        $path = rawurldecode(Str::after($url, $containerUrl));
```

We are running that change as a local subclass of the repository and it resolves all affected assets, with no other behaviour changing.

DonnyOexman:
Fixes #609.

`Statamic\Eloquent\Assets\AssetRepository::findByUrl()` passes the path straight into the database query without decoding it, unlike `Statamic\Assets\AssetRepository::findByUrl()` in core, which does:

```php
$path = rawurldecode(Str::after($url, $containerUrl));
```

Asset URLs are percent-encoded, so any filename containing a space, parentheses, `@`, a comma or an accented character produces a path like `my%20file%20%281%29.jpg`. That never matches the stored `basename`, and `findByUrl()` returns `null` for an asset that exists.

This is easy to hit in practice: the returned `null` reaches `ImageGenerator::generateByAsset()` via the `glide` tag, which has no null check and throws `Call to a member function isVideo() on null`. Because that's an `\Error` and `Glide::generate()` only catches `\Exception`, it escapes the tag's own error handling and 500s the whole page rather than skipping the image. SEO Pro renders `og:image` through that path, so one asset with a space in its name can take down every page referencing it.

### The change

Restores the `rawurldecode()` call so the Eloquent driver matches core.

### Tests

Added `tests/Repositories/AssetRepositoryTest.php` with two cases: a plain filename (passes before and after, guards against regressions) and a percent-encoded one (fails with `Call to a member function id() on null` before the change, passes after).

The two pre-existing failures in `tests/Entries/EntryRepositoryTest.php` are also present on a clean `5.x` checkout and are unrelated to this change.

### DIFF
--- a/src/Assets/AssetRepository.php
+++ b/src/Assets/AssetRepository.php
@@ -53,7 +53,7 @@ class AssetRepository extends BaseRepository
             $url = $siteUrl.$url;
         }
 
-        $path = Str::after($url, $containerUrl);
+        $path = rawurldecode(Str::after($url, $containerUrl));
 
         if (Str::startsWith($path, '/')) {
             $path = substr($path, 1);
--- a/tests/Repositories/AssetRepositoryTest.php
+++ b/tests/Repositories/AssetRepositoryTest.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Tests\Repositories;
+
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Storage;
+use PHPUnit\Framework\Attributes\Test;
+use Statamic\Facades;
+use Tests\TestCase;
+
+class AssetRepositoryTest extends TestCase
+{
+    use RefreshDatabase;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        config(['filesystems.disks.test' => [
+            'driver' => 'local',
+            'root' => __DIR__.'/tmp',
+        ]]);
+
+        Storage::fake('test', ['url' => '/assets']);
+
+        tap(Facades\AssetContainer::make('test')->disk('test'))->save();
+    }
+
+    protected function tearDown(): void
+    {
+        app('files')->deleteDirectory(__DIR__.'/tmp');
+
+        parent::tearDown();
+    }
+
+    #[Test]
+    public function it_finds_an_asset_by_url()
+    {
+        Storage::disk('test')->put('a.jpg', '');
+        $asset = tap(Facades\Asset::make()->container('test')->path('a.jpg'))->save();
+
+        $this->assertEquals($asset->id(), Facades\Asset::findByUrl($asset->url())->id());
+    }
+
+    /**
+     * The URL is written out encoded rather than taken from $asset->url(), because
+     * whether that method encodes has varied across Statamic versions. What matters
+     * here is that an encoded URL resolves, however it was produced.
+     */
+    #[Test]
+    public function it_finds_an_asset_by_url_when_the_path_is_percent_encoded()
+    {
+        Storage::disk('test')->put('my file (1) @ café.jpg', '');
+        $asset = tap(Facades\Asset::make()->container('test')->path('my file (1) @ café.jpg'))->save();
+
+        $this->assertEquals(
+            $asset->id(),
+            Facades\Asset::findByUrl('/assets/my%20file%20%281%29%20%40%20caf%C3%A9.jpg')->id()
+        );
+    }
+}


DonnyOexman:
Heads up on the red CI here: it's pre-existing on `5.x` and not caused by this change.

The nightly runs on `5.x` have been failing since 26 Aug, with no commits to the branch since 11 Aug (#604):

- 25 Aug, green: https://github.com/statamic/eloquent-driver/actions/runs/32792258886
- 26 Aug, red: https://github.com/statamic/eloquent-driver/actions/runs/32913725839
- red every night since: https://github.com/statamic/eloquent-driver/actions/workflows/tests.yml?query=branch%3A5.x+event%3Aschedule

Both failures are `EntryRepositoryTest::it_updates_the_order_of_all_entries_in_a_collection` and `::it_updates_the_order_of_specific_entries_in_a_collection`, and they bisect to statamic/cms v6.29.0, released 25 Aug. Saving an entry into a structured collection now appends it to the structure tree, so `Entry::order()` returns its tree position instead of the default. Immediately after the four saves the tree already holds those entries, so the guard assertion sees `1, 2, 3, 4` where it expects `1, 1, 1, 1`. Pinning `statamic/cms` to 6.28.0 makes both pass again.

The same split is visible in this PR's own matrix: the `prefer-lowest` jobs resolve cms v6.24.2 and run the entire suite green — `OK, but some tests were skipped! Tests: 392, Assertions: 1510, Skipped: 23` — and are marked red only because fail-fast cancelled them once a `prefer-stable` job had failed.

The two tests added here pass on both v6.24.2 and v6.30.0.

Happy to open a separate PR adjusting those two `EntryRepositoryTest` assertions if the new core behaviour is intended, rather than bundling it in here.

DonnyOexman:
Tracked down where this diverged, which might make the change easier to review.

statamic/cms#15146 ("Encode asset URLs", released in v6.29.0) changed two things in one go: `Asset::url()` and `absoluteUrl()` started percent-encoding the path via a new `encodedPath()`, and `Statamic\Assets\AssetRepository::findByUrl()` got a matching `rawurldecode()` so the round trip stayed intact.

https://github.com/statamic/cms/commit/07ad17fc2608ba82be938d41591ea26dec81f230

`Statamic\Eloquent\Assets\AssetRepository` overrides `findByUrl()` to resolve through `findById()` instead of `$container->asset()`, so it picked up the encoding through the inherited `Asset::url()` but not the decode that was added alongside it. That's the whole bug, and it dates the regression precisely: on v6.28.0 and earlier nothing encoded the path so the lookup matched, from v6.29.0 onward it doesn't.

This PR just applies the same `rawurldecode()` to the override.

duncanmcclean:
> Heads up on the red CI here: it's pre-existing on 5.x and not caused by this change.

Should be fixed now. I've just merged #611 and merged `5.x` into your branch. 🙂

daun:
Can confirm both the original issue and this fix to be working 🧃

PR #613, opened 2026-09-02
Store `publish_at` on revisions

edalzell:
Companion to statamic/cms#11508 (scheduled revision publishing), which adds a `publish_at` date to revisions. This persists it in the `revisions` table so the core scheduler can query `publish_at` against the eloquent driver.

Adds a nullable `publish_at` timestamp column (fresh installs via the create migration, existing installs via an update script that publishes `add_publish_at_to_revisions_table`), maps it in `Revision::fromModel()`/`toModel()`, exposes it as a real column in `RevisionQueryBuilder` so `whereNotNull('publish_at')` / `where('publish_at', '<=', now())` hit the column rather than `attributes->`, and round-trips it in the import/export commands.

Depends on statamic/cms#11508 for `Revision::publishAt()`, so tests will fail until that ships and the `statamic/cms` constraint is bumped.

### DIFF
--- a/database/migrations/2024_03_07_100000_create_revisions_table.php
+++ b/database/migrations/2024_03_07_100000_create_revisions_table.php
@@ -14,6 +14,7 @@ return new class extends Migration
             $table->string('action')->index();
             $table->string('user')->nullable();
             $table->text('message')->nullable();
+            $table->timestamp('publish_at')->nullable();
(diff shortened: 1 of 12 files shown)


duncanmcclean:
Converting to a draft until the Core PR is merged.

